PR #1, opened 2026-07-12
feat: Claude rate-limit session handoff

williamttruong:
## Summary
Adds deterministic Claude session-limit handoff support so a hard stop can leave a resume note for the next chat without requiring AI after the model dies.

## What changed
- Claude `StopFailure` / failed host stops now preserve structured error codes such as `rate_limit`
- Knowl stores a project-local `pending_handoff` state item from the last checkpoint + failure metadata
- Next `SessionStart` injects that handoff first, then normal recent context, and archives it so it is delivered once
- Non-limit failures still record a lower-urgency handoff
- Tests cover host-hook normalization, handoff detection, and lifecycle store/inject/consume flow
- Version bump to `1.1.1`

## Why
Claude session limits currently kill a turn without a durable resume note. Structured Knowl writes do not need AI, so this path can still save a handoff even when the model is already dead.

## Test plan
- [x] `vitest run tests/store/host-lifecycle.test.ts tests/store/session-handoff.test.ts tests/cli/host-hook.test.ts`
- [ ] Manual Claude project with hooks installed: force/stop-fail with rate limit, restart session, confirm handoff context appears once


dat999zx:
reviewed, upgraded at host-neutral-hard-stop-handoff

PR #7, opened 2026-08-01
feat(transcript-search): searchable session transcripts as a lossless floor under the atoms

williamttruong:
## What this adds

Two MCP tools — `knowl_transcript_search` and `knowl_transcript_read` — that make raw Claude Code session transcripts searchable from Knowl.

## Why

Knowl stores distilled atoms. That is lossy by construction: whatever the writer did not judge salient is gone, and there is no fallback — the documented weakness of extract-and-store memory systems. The `.jsonl` transcripts are the lossless record underneath, and they were unreachable. This turns a memory miss into a slower lookup instead of amnesia.

It also closes a gap in promotion. `candidate-promotion.ts` promotes hook-captured session events; a fact recovered by reading history had no path back into memory, so the same lookup repeated every session. Each hit carries a `transcript://<sessionId>#L<line>` locator, and the **result** — not only the tool description — instructs the caller to promote what it used. Descriptions are read once; results are read every time.

## Design notes

**FTS5, not a hand-rolled index.** The first implementation built its own postings table at ~74 rows per message. Measured: three 8s passes covered 14MB of a 930MB archive, so a cold index would have taken ~30 minutes. FTS5 ships BM25, is already used here for `knowledge_items`, and with `content='transcript_messages'` avoids storing text twice. The same budget then covered 12–15k messages per pass — roughly 20× faster.

**Field weighting on top of BM25.** `bm25()` returns negative-is-better; it is inverted and multiplied by a stored weight — user 2.0, assistant 1.0, and 0.3 for a message more than half tool output, so a pasted file cannot win on volume. The candidate window is 4× the requested depth because re-weighting can promote a row FTS5 ranked lower.

**`sessionId` scoping.** A handoff brief that names a parent session needs a way to ask that transcript specifically before widening to the whole archive. Accepts a full id or a unique prefix.

**Incremental by byte offset.** Transcripts are append-only, so `transcript_files.bytes_indexed` is a valid resume point.

**Bounded latency.** Indexing runs inside a tool call, so it has an 8s budget and resumes mid-file. A partial index reports `INDEX STILL WARMING` in the result, because a caller told nothing would read "no matches" as proof of absence.

**Small write batches.** Long write transactions starve a concurrently served session into `SQLITE_BUSY`. Hit in practice, not anticipated.

**Concurrent serve processes are serialized per file.** Knowl runs one serve process per connected session, so two sessions on one project index concurrently. A per-file lease (compare-and-set on the watermark stamp, 20s expiry) keeps them off each other's files; a cross-file id collision aborts that file's pass and defers to the next one, and a claimant purges any rows a crashed predecessor committed past the watermark, embeddings included — a purged id can be reallocated, and a stale cached vector would otherwise attach to the wrong message.

**Optional hybrid ranking.** When vector search is enabled, top candidates are re-ranked and the two orders fused by Reciprocal Rank Fusion (k=60), combining *positions* rather than scores, since bm25 magnitudes and cosine similarities are not on a comparable scale. Vectors are cached per message. A disabled provider or missing model degrades to lexical ranking rather than failing.

## Limitations, stated rather than glossed

- **Semantic recall is bounded by the lexical pass** — only the top 50 BM25 candidates are embedded. Embedding an entire archive was not worth the cost, which is why the lexical side is tuned to stand alone.
- **In-place edits at identical file size go undetected.** A file that *shrinks* is treated as rewritten and rebuilt; one edited without changing size is not. Append-only is an assumption, not something enforced.
- **Old-format archives that interleave subagent turns** (`isSidechain` entries inside the parent transcript) would have those indexed as ordinary parent-session messages. Verified absent in a current-format 66-session archive; disclosed rather than special-cased.
- **Index storage is not reclaimed** when transcripts are deleted from disk; rows for a vanished session persist until that session is rebuilt.
- **The compact guidance card grew** to 1,871 / 1,922 chars (468 / 481 tokens) against its 2,000-char and 500-token ceilings. Still inside both, with less headroom than before — worth knowing before the next tool is added.

## Testing

16 tests in `tests/store/transcript-search.test.ts`: ranking order, noise exclusion, tool-output weighting, `sessionId` scoping by id and prefix, incremental append, rewrite-triggered rebuild, budget interruption with resume and no double-indexing, lease deferral and stale-lease reclaim, embedding purge on rebuild, RRF fusion with a stub embedder, and graceful degradation when the embedder throws.

Registered in the canonical inventory, so the guidance contract tests, README table, and `tools/list` all stay aligned.

Verified end-to-end over the real MCP stdio protocol against a 66-session / ~930MB archive: 53,744 messages indexed across five passes, then **21–29 ms per query** once warm.

Note for reviewers: the full suite is load-sensitive on a busy machine — `query-command`, `workspace-query` and `foreign-item-refusal` can time out under a 135-file parallel run and pass in isolation. Verified identical behaviour on pristine `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -963,6 +963,8 @@ Knowl exposes exactly 24 MCP tools:
 | `knowl_recent` | Compact recent context when lifecycle bootstrap is unavailable or a refresh is needed |
 | `knowl_state` | Broad active-memory status or hierarchical project summary |
 | `knowl_context` | Compose an explicitly token-budgeted local context pack |
+| `knowl_transcript_search` | Search the verbatim session transcripts when memory misses or is ambiguous |
+| `knowl_transcript_read` | Read one transcript entry in full when a snippet cannot settle it |
 | `knowl_task_start` | Start one manual work loop when verified lifecycle hooks are unavailable |
 | `knowl_task_checkpoint` | Checkpoint meaningful manual-loop progress or blockers |
 | `knowl_task_finish` | Finish one manual work loop after verification |
--- a/src/core/knowl-guidance.ts
+++ b/src/core/knowl-guidance.ts
@@ -13,6 +13,11 @@ export const KNOWL_MCP_TOOL_GROUPS = [
     routing: 'Use recent only without lifecycle bootstrap or for an explicit refresh; state for broad status; context for an explicitly token-budgeted pack.',
   },
   {
+    label: 'Verbatim transcripts',
+    tools: ['knowl_transcript_search', 'knowl_transcript_read'],
+    routing: 'The lossless record under the distilled atoms. Use only after knowl_query misses or returns something stale, ambiguous, or lower-confidence than the question needs, and use it instead of guessing about a past decision. Scope with sessionId when a handoff names a parent session, then widen. Read one entry in full only when a snippet cannot settle it. Promote what you used with knowl_store or knowl_update citing the returned locator, so the same lookup is never paid for twice.',
+  },
+  {
     label: 'Manual work loop',
     tools: ['knowl_task_start', 'knowl_task_checkpoint', 'knowl_task_finish'],
     routing: 'Use only without verified lifecycle hooks: start once, checkpoint meaningful milestones or blockers, and finish once after verification.',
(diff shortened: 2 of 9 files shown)


dat999zx:
Superseded by #12, which reimplements this against current `main`.

The core argument here was right and #12 keeps it: distilled atoms are lossy by construction, the `.jsonl` transcripts are the lossless record underneath, and reaching them turns a memory miss into a slower lookup instead of amnesia. Also kept: FTS5 over a hand-rolled index, RRF fusion, and byte-offset resume.

What changed, and why:

- **Off by default.** The compact guidance card is a token cost every session of every user pays. #12 registers nothing unless `search.transcripts.enabled` is set.
- **Prose only.** Measured on this repo's archive: prose is 2.7% of 80.9 MB. Skipping tool output shrinks the corpus ~37x and removes the 0.3 weight this PR needed to stop pasted files winning on volume.
- **Pointers, not text.** Rows are `(session, line, role)`; bodies stay in the `.jsonl`. Under 3 MB of index instead of a second copy of the archive.
- **Its own database.** `.knowl/transcripts.db`, so a backfill cannot contend with the live session and "off" means a file that does not exist.

One factual correction worth recording: this PR's stated limitation says subagent turns are either absent or interleaved as `isSidechain` entries in the parent file. In the current format they are separate files under `<session-uuid>/subagents/`. Top-level-only enumeration finds 24 of 76 files here.

PR #8, opened 2026-08-01
feat(session-directory): browsable session inventory (stacked on #7)

williamttruong:
> **Stacked on #7** (`feat/transcript-search`) — this branch contains #7's commit plus one more. Review only the top commit here; merge #7 first, after which this shows only the session-directory change.

## What this adds

`knowl_session_list`: a browsable, agent-facing inventory of a project's Claude Code sessions — best-known name, opening ask, activity, and the knowledge each session promoted into memory. It answers "which session was about X?" and "resume or start fresh?", then hands its `sessionId` to `knowl_transcript_search` to read into the chosen session.

## Research before build

Nine shipped tools surveyed first (native `/resume` picker, claude-historian-mcp, claude-session-manager, fork-aware claude-session-mcp, iannuttall/claude-sessions, ClaudeHistoryMCP, tim0120's, raine/claude-history, Session Continuity MCP). Cross-cut findings that shaped this design:

- Everyone enumerates from the transcript store; **nobody reads the transcript's own title entries** — external tools fall back to filenames or first prompts, while a `custom-title` from a user rename is sitting in the file.
- **Deterministic summaries are the norm**; AI summaries are the exception and require a model. This tool is deterministic throughout.
- **Nobody models what a session was for.** The closest attempt is manual session-documentation commands whose own README names forgetting as the failure mode.
- The native picker's issue tracker (anthropics/claude-code #23375, #24435, #25130, #29052, #35698) documents the pain directly: 50-session cap, ~10 enriched, unnamed sessions filtered out entirely.

Skipped deliberately: AI summarization (provider-dependent), regex "decisions/solutions" extraction (knowl's hook-driven candidate promotion already owns that signal), message-level fork trees (solved elsewhere).

## How it works

The transcript indexer from #7 already streams every line; it now also captures — at zero additional IO — the session's title entries (`custom-title` > `agent-name` > `ai-title`, later beats earlier within a rank, carried across incremental passes and reset when a file is rewritten) and the first real user ask as a one-line descriptor. `knowl_session_list` joins that with per-session activity and, through `host_session_bindings → memory_sessions → promotion_items`, the titles of atoms each session promoted — the one enrichment a plain transcript viewer cannot offer. Keyword filter over intent fields only; no session cap; unnamed sessions included.

Schema: three columns on `transcript_files` (unreleased table from #7), with a self-heal that resets watermarks so pre-existing dev indexes refill names on the next pass.

## Status and cards — the gap nobody in the survey fills

Each entry carries a `status`, derived at read time and never stored: `interrupted` when an unconsumed crash handoff names the session (including pre-tag handoffs that carry the id only in content JSON — real archives have these), `active` on a recent bound heartbeat or transcript write, `idle` otherwise. Fine-grained working/stalled tiers from live monitors were considered and rejected — their precondition (real-time observation) is absent in a recall surface.

Sessions can also declare a card — any atom tagged `session-card` + `session:<id>` via the existing `knowl_store`, newest wins — and the directory surfaces it beside the derived status, included in keyword matching. Intent declaration costs zero new tools.

## Measured

Against a 66-session / ~930MB archive: `query: "ui"` returned the owner's main UI-design session and four spawned UI worksessions, correctly named, in one call — the exact question that motivated the feature.

## Testing

7 new tests (rename-beats-generated-title, unnamed-session inclusion, keyword filter, incremental rename pickup, lifecycle-chain enrichment, status derivation incl. legacy content-only handoffs, card surfacing + filtering) plus the four-part guidance contract updated: 27 tools, compact card measured at 1,917/1,968 chars against the 2,000 ceiling — headroom is thinning; worth a thought before the next tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -963,6 +963,7 @@ Knowl exposes exactly 24 MCP tools:
 | `knowl_recent` | Compact recent context when lifecycle bootstrap is unavailable or a refresh is needed |
 | `knowl_state` | Broad active-memory status or hierarchical project summary |
 | `knowl_context` | Compose an explicitly token-budgeted local context pack |
+| `knowl_session_list` | Browse past sessions as an inventory: names, opening asks, promoted knowledge |
 | `knowl_transcript_search` | Search the verbatim session transcripts when memory misses or is ambiguous |
 | `knowl_transcript_read` | Read one transcript entry in full when a snippet cannot settle it |
 | `knowl_task_start` | Start one manual work loop when verified lifecycle hooks are unavailable |
--- a/README.md
+++ b/README.md
@@ -963,6 +963,8 @@ Knowl exposes exactly 24 MCP tools:
 | `knowl_recent` | Compact recent context when lifecycle bootstrap is unavailable or a refresh is needed |
 | `knowl_state` | Broad active-memory status or hierarchical project summary |
 | `knowl_context` | Compose an explicitly token-budgeted local context pack |
+| `knowl_transcript_search` | Search the verbatim session transcripts when memory misses or is ambiguous |
+| `knowl_transcript_read` | Read one transcript entry in full when a snippet cannot settle it |
 | `knowl_task_start` | Start one manual work loop when verified lifecycle hooks are unavailable |
 | `knowl_task_checkpoint` | Checkpoint meaningful manual-loop progress or blockers |
 | `knowl_task_finish` | Finish one manual work loop after verification |
--- a/src/core/knowl-guidance.ts
+++ b/src/core/knowl-guidance.ts
@@ -13,6 +13,11 @@ export const KNOWL_MCP_TOOL_GROUPS = [
     routing: 'Use recent only without lifecycle bootstrap or for an explicit refresh; state for broad status; context for an explicitly token-budgeted pack.',
   },
   {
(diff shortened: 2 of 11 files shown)


williamttruong:
The failing check here is the same-millisecond access-record ordering flake fixed in #9 (random-hex id as tiebreak → coin-flip order on fast runners; [this run](https://github.com/dat999zx/knowl/actions/runs/30705760305) hit tails). Suggested merge order: #9 → #7 → #8.

dat999zx:
Superseded by #12, which ports this onto the transcript index that shipped there.

The research behind this is the part that carried over, and it held up: no surveyed tool reads the transcript's own title entries, so they fall back to filenames or first prompts while a user's `custom-title` sits unread in the file. #12 keeps the precedence rule (rename > agent name > generated title, later wins within a rank but never across), the no-session-cap decision, unnamed sessions described by their opening ask, and the promoted-knowledge join through `host_session_bindings → memory_sessions → promotion_items`.

What could not carry over: this was stacked on #7, whose index #12 replaced, so `session-directory.ts` reads columns that no longer exist and ships its own copies of discovery, indexing and search. #12 adds the three columns to the shipped schema and reuses `src/transcripts/` rather than running a second index beside it.

PR #9, opened 2026-08-01
fix(access): stable ordering for same-millisecond access records

williamttruong:
One line: `ORDER BY retrieved_at ASC, id ASC` → `rowid ASC`. Ids are random hex, so two access records written in the same millisecond come back in coin-flip order; the access-feedback test asserts insertion order and flakes on fast runners — observed on an unrelated PR's CI run ([this failure](https://github.com/dat999zx/knowl/actions/runs/30695537241)). rowid is monotonic and therefore insertion order.

### DIFF
--- a/src/store/access-feedback.ts
+++ b/src/store/access-feedback.ts
@@ -88,7 +88,7 @@ export async function recordKnowledgeFeedback(input: Omit<KnowledgeAccessInput, 
 
 export async function listKnowledgeAccess(itemId: string): Promise<KnowledgeAccess[]> {
   const result = await getClient().execute({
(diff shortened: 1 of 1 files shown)


dat999zx:
Closing unmerged: this fix is already on `main`.

`src/store/access-feedback.ts:95` reads `ORDER BY retrieved_at ASC, rowid ASC` today — byte-identical to what this PR proposes. It arrived by another route, so there is nothing left to merge.

The diagnosis was right: ids are random hex, so same-millisecond records came back in coin-flip order and the access-feedback test flaked on fast runners. Thanks for tracking it down.

PR #10, opened 2026-08-02
feat(handoff): park a workstream for the next session on purpose

williamttruong:
Parking a workstream and crashing out of one are different events, and the handoff a session receives should say which happened.

`knowl_handoff` writes a deliberate baton: goal, what is done, the next action, any blocker, and whether the work was verified. The next session started in that project receives it once and it is then archived — a one-shot pass, not a durable note. Durable facts still belong in `knowl_store`.

**Why the kind matters.** The existing crash handoff opens with "ended before a clean finish", which is exactly wrong for work someone parked on purpose — it invites the next session to treat a clean stopping point as damage to repair. `SESSION_HANDOFF_KINDS` is now a single `const` array that the type, the writer and the parser all derive from, so a new kind cannot be written-but-unreadable. That was a real defect: the parser had its own hardcoded allowlist, so any kind added to the type would have been persisted and then silently dropped on read.

One baton per project at a time — writing another replaces it.


dat999zx:
Superseded by #12, which rebases this onto current `main`.

The central insight is kept intact: a parked workstream and a crash deserve opposite openings, and telling a session that stopped cleanly it "ended before a clean finish" invites it to go looking for damage that does not exist.

Also kept, and worth calling out — this PR identified a real defect independently of the feature. `session-handoff.ts` validated the parsed kind against a hardcoded array while `SessionFailureKind` was declared separately, so any kind added to the type was written successfully and then dropped on every read, with nothing to notice at write time. #12 fixes that **before** adding the new kind, because doing it after produces a feature that silently does not work.

The conflict-key normalization fix travels with it.

Not carried over: the `3.0.0-fork.1` version bump, which is fork-sync bookkeeping that means nothing in this repo.

PR #11, opened 2026-08-02
feat(transcript-search): rank the whole archive semantically, not BM25's top 50 (stacked on #10)

williamttruong:
Stacked on #7, #8 and #10 — review the last two commits.

## The bound this removes

Semantic search re-ranked BM25's top 50. That means it could reorder keyword results but never reach past them, so the query the feature exists to serve was structurally unreachable: **the one where the words you remember are not the words that were used.** That message is never in the shortlist, and no embedding model can fix a candidate set it never sees.

Every indexed message now carries a vector. The lexical and semantic orders are built independently over the whole archive and fused with RRF, so a message sharing *no word* with the query can win outright. There is a test for exactly that, and it fails against the old design.

## Why int8 and not the published answer

Storing float32 for a 69k-message archive is 106 MB to read and multiply per query, so the representation has to shrink. The [sentence-transformers guidance](https://sbert.net/examples/sentence_transformer/applications/embedding-quantization/README.html) says binary quantization retains ~96% of retrieval quality with a rescoring pass, so I measured it here — 350 project atoms, 17 recall queries, float32 reference MRR 0.662:

| | MRR | size (69k msgs) | full scan |
|---|---|---|---|
| float32 | 0.662 | 106 MB | 91 ms |
| **int8** | **0.668** | **27 MB** | **36 ms** |
| binary | 0.310 | 3.3 MB | 5 ms |
| binary + float32 rescore | 0.664 | 110 MB | — |

Binary collapsed. The published figures come from 1024-dim models; at 384 dims one sign bit per dimension is not enough information to hold the ranking. Recovering it needs a float32 rescoring pass, which means storing the float32 vectors *as well* — 4x the disk to undo damage the compression caused. int8 measured no loss at a quarter the size with no rescoring stage at all.

The scale is `6 / sqrt(dims)` rather than a constant: L2-normalised components have RMS `1/sqrt(dims)`, so this clips at ~6 sigma and adapts to any model's dimensionality. For 384-dim bge-small it gives 0.306, against a measured largest component of 0.327 and p99.9 of 0.262.

Brute force rather than ANN, on purpose: 36 ms needs no index, and an index costs rebuilds, recall knobs, and a native extension `@libsql/client` cannot load anyway.

## Operationally

- `knowl reindex --transcripts [--budget <minutes>]` backfills. Resumable by construction — "messages with no vector" *is* the resume state, so an interrupted pass leaves no bookkeeping to repair.
- Searches top up ~1.5s worth, newest first, so an archive warms as it is used and the current session's messages are covered first.
- The MCP footer reports coverage as `embedded/indexed`. "BM25 + semantic" over 2% of an archive is a different claim from the same words over all of it, and only one of them justifies trusting a near-miss result.
- Vectors for a superseded model are dropped on the next pass. There is now one vector per *message* rather than per re-ranked candidate, so a stale model's copy is a full dead duplicate of the archive.

## Three concurrency bugs, each found by running a real backfill

None of these were theoretical — a 41k-message job beside a live session hit all three.

1. **Bootstrap had no retry on a contended open.** Opening a database runs migrations and DDL — real writes — against a database other sessions are also writing to. The whole command died on a transient lock.
2. **No `busy_timeout` anywhere.** SQLite's default busy handler gives up instantly. Survivable while indexing ran in 8-second slices inside the process holding the lock; not survivable for anything longer. Set at the connection, so statements nobody thought to wrap are covered too.
3. **A connection pinned to a stale read snapshot can never commit again.** `SQLITE_BUSY_SNAPSHOT` is permanent for that connection, so backing off waits for a condition only a reconnect clears. Measured directly: a fresh process wrote the same database in **71 ms** while the long-lived job had been failing on it for **fourteen minutes**. The backfill now reconnects and resumes.

Point 3 is also why the per-batch progress count had to go: it was a `LEFT JOIN` over every message in the project, running after each batch of 16, interleaved with the writes it was reporting on.

---

## Update: the model underneath this changed, and one number above is superseded

**Model.** `bge-small-en-v1.5` → `granite-embedding-small-english-r2`, measured the same way, all rows from one session on one corpus snapshot:

| model | pooling | @1 | @3 | MRR |
|---|---|---|---|---|
| **granite-embedding-small-english-r2** | **CLS** | **59%** | **94%** | **0.750** |
| bge-small-en-v1.5 | mean | 53% | 65% | 0.621 |
| granite-embedding-97m-multilingual-r2 | CLS | 47% | 76% | 0.643 |
| granite-embedding-97m-multilingual-r2 | mean | 18% | 29% | 0.337 |

The multilingual model of triple the size is not better on an English corpus. The last row is the same weights as the row above it with the wrong pooling — which is why pooling now follows the model family instead of being hardcoded, and why that has tests.

**Correction.** An earlier commit message here cites **61 ms/doc** for embedding, taken from the atom corpus (~2,000 chars/doc). Real transcript messages have a median of **169 characters**, and the measured production rate is **42 ms/doc** for bge-small and **60 ms/doc** for granite. The 61 figure was never wrong for what it measured; it was the wrong corpus to quote for this workload. Full 58k-message backfill: ~48 min then, ~54 min now.

**Vector storage is unaffected** — Granite is also 384 dimensions, so every int8 quantization measurement above still holds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -963,6 +963,7 @@ Knowl exposes exactly 24 MCP tools:
 | `knowl_recent` | Compact recent context when lifecycle bootstrap is unavailable or a refresh is needed |
 | `knowl_state` | Broad active-memory status or hierarchical project summary |
 | `knowl_context` | Compose an explicitly token-budgeted local context pack |
+| `knowl_session_list` | Browse past sessions as an inventory: names, opening asks, promoted knowledge |
 | `knowl_transcript_search` | Search the verbatim session transcripts when memory misses or is ambiguous |
 | `knowl_transcript_read` | Read one transcript entry in full when a snippet cannot settle it |
 | `knowl_task_start` | Start one manual work loop when verified lifecycle hooks are unavailable |
--- a/README.md
+++ b/README.md
@@ -966,6 +966,7 @@ Knowl exposes exactly 24 MCP tools:
 | `knowl_session_list` | Browse past sessions as an inventory: names, opening asks, promoted knowledge |
 | `knowl_transcript_search` | Search the verbatim session transcripts when memory misses or is ambiguous |
 | `knowl_transcript_read` | Read one transcript entry in full when a snippet cannot settle it |
+| `knowl_handoff` | Park the current workstream; the next session in this project receives it once |
 | `knowl_task_start` | Start one manual work loop when verified lifecycle hooks are unavailable |
 | `knowl_task_checkpoint` | Checkpoint meaningful manual-loop progress or blockers |
 | `knowl_task_finish` | Finish one manual work loop after verification |
--- a/README.md
+++ b/README.md
@@ -963,6 +963,8 @@ Knowl exposes exactly 24 MCP tools:
 | `knowl_recent` | Compact recent context when lifecycle bootstrap is unavailable or a refresh is needed |
 | `knowl_state` | Broad active-memory status or hierarchical project summary |
 | `knowl_context` | Compose an explicitly token-budgeted local context pack |
+| `knowl_transcript_search` | Search the verbatim session transcripts when memory misses or is ambiguous |
+| `knowl_transcript_read` | Read one transcript entry in full when a snippet cannot settle it |
 | `knowl_task_start` | Start one manual work loop when verified lifecycle hooks are unavailable |
 | `knowl_task_checkpoint` | Checkpoint meaningful manual-loop progress or blockers |
 | `knowl_task_finish` | Finish one manual work loop after verification |
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dat999zx/knowl",
-  "version": "2.11.1",
+  "version": "3.0.0-fork.1",
   "description": "A Knowledge Operating System for AI Agents",
   "main": "dist/index.js",
   "type": "module",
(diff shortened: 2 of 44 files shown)


dat999zx:
Superseded by #12. Most of this had already landed by other routes; the rest is rebuilt there.

**Already on `main`:** the embedding model work. `granite-small-en-r2` is the default preset, chosen the way this PR argued for — by measurement rather than leaderboard, with pooling following the model family because a wrong value produces plausible vectors that rank badly with nothing to notice at runtime.

**Rebuilt in #12:** whole-corpus semantic ranking. The bound this PR removed was the right one to remove — re-ranking BM25's top 50 can reorder keyword results but never reach past them, so the query the feature exists for is structurally unreachable. #12 keeps that and the int8 measurements (float32 MRR 0.662 / 106 MB, int8 0.668 / 27 MB, binary 0.310 — binary collapses at 384 dims). Also kept: the three concurrency findings, each of which cost real debugging here. `SQLITE_BUSY_SNAPSHOT` being permanent for a pinned connection, so a reconnect is the only fix, is the one that would have been hardest to rediscover.

**Also rebuilt:** `knowl_park` / `knowl_resume`. One correction to an assumption made while triaging: this is **not** redundant with #10's baton. One is a single baton per project, pushed to the next session and spent on use; the other is many parked workstreams, pulled by a key the user holds, from any directory, repeatedly. Both shipped.

Not carried over: the `project-dir` refactor and the `freshness`/`gc`/`host-lifecycle` churn, which are unrelated to either feature.

PR #13, opened 2026-08-03
fix(serve): complete the MCP handshake before opening the database

williamttruong:
`startMcpServer` opens the database, runs `bootstrapSchema` and looks the project up **before** it connects the transport. The host allows a fixed window for the handshake — 30s in Claude Code, configurable with `MCP_TIMEOUT` — and kills the process when it expires. So every one of those steps is charged against a deadline none of them knows about.

## Why this is worth fixing

None of that work is needed to answer `initialize`. The tool list is a static literal, the declared capabilities are constants, and the resource list is static too. The database is needed by tool *calls*, and those answer to the tool-call timeout instead — hours rather than seconds, which is the right clock for work whose duration depends on what other processes are doing to the same file.

And `bootstrapSchema` is not a cheap open. Every read-write open runs the base statements, `migrateLegacyProjectSchema`, the schema statements, ten `ensure*Columns` passes, `backfillKnowledgeAssertions` and `repairSkillForeignKeys`. That is real write work against a file other Knowl processes may be opening at the same moment — one per connected session, plus anything started from a terminal. It is usually fast and occasionally not, and when it is not, the whole server dies rather than that one open merely being slow.

The failure is silent in a particularly unhelpful way: the host kills the process, and the only thing in its log is the boot banner, because the line printed after the database opens never runs. There is nothing to distinguish "hung opening the database" from "crashed on startup".

## How I hit it

In my host's MCP logs there are 22 kills of exactly this shape — `connection timed out after 30000ms`, stderr ending at the boot banner — across four projects, sometimes several within the same second, which is the signature of processes contending rather than of any one project being slow.

**Being straight about the conditions:** I run a fork that adds `busy_timeout` and a retry loop to `acquireClient`, which makes a contended open wait rather than fail, and that is what turns this from "occasionally slow" into "reliably fatal" on a busy machine. This repo has neither, so the exposure here is smaller. But the structural problem — unbounded work sitting on a bounded deadline — is the same in this codebase, and a slow open for any reason (large migration, slow or network-backed disk, a machine under load) lands on the connect deadline today.

## The change

**Connect first, initialize behind it.** Tool calls and resource reads await readiness. A failure still surfaces exactly as before: the error is captured and rendered through `formatInitError`.

The wait is awaited in three places rather than one — inside `callTool`, in the `CallToolRequest` wrapper *before* it reads `projectRoot` for the change watermark, and in `ReadResource`, which would otherwise see neither a project id nor an init error and proceed against `null`. `createMcpServer` keeps its positional signature and gains an optional `DeferredServerState`, so every existing caller — the tests among them — is unchanged.

The banner now prints straight after the handshake rather than after initialization, so a stalled boot is distinguishable from a dead one in the host log.

**Plus a startup trace,** because this took far longer to diagnose than it should have. A watchdog fires at 5s/15s/25s — inside the host's window — and names the phase currently running, since a process killed at the deadline never reaches its own summary line. It writes to stderr (landing in the host's MCP log beside the host's own timeout line) and to a machine-wide JSONL that survives the kill and spans projects. It is silent on healthy boots. `knowl diagnose-startup [--since <hours>]` reads it back: boots started vs became ready, per-phase percentiles, model load times split cold/warm, stalls by owning phase, and concurrent-boot bursts. `KNOWL_DISABLE_STARTUP_TRACE=1` turns it off.

Model loads are timed because "the embedding model is what makes startup slow" is the intuitive explanation and it was worth being able to settle rather than argue about — `serve` never builds a pipeline during startup, and the trace now reports whether any load landed on a startup path at all.

## Measured

| | before | after |
|---|---|---|
| handshake, project A | blocked behind the database open | **666ms** |
| handshake, project B | blocked behind the database open | **766ms** |
| database open alone | 111ms – 2,066ms, **on the deadline** | unchanged, **off the deadline** |

The slow opens are unchanged. They are simply no longer able to kill the server.

## A deeper cause this does NOT fix

This moves the wait off the fatal path; it does not remove it. The wait exists because `bootstrapSchema` re-runs the full migration suite on every read-write open, taking write locks on a shared file.

`stampSchemaVersion` already has the right idea and says so in its own comment — *"Every rw open runs this. Writing unconditionally means several processes racing to bootstrap the same file each take a header-write lock even when nothing changed"* — but the fast path is applied to that one statement rather than to the suite. Gating the whole of `bootstrapSchema` behind a cheap check would turn ~15 write operations into one read on the hot path. `KNOWL_SCHEMA_VERSION` can't serve as that check on its own, since additive columns deliberately don't bump it, so it would need a schema fingerprint — the shape `ensureEmbeddingProfileColumns` already uses for the embedding profile. Happy to follow up with that separately if you'd want it.

## Testing

- Full suite on this branch: **1340 passed, 1 failed**. The failure is `tests/transcripts/backfill.test.ts` — a wall-clock budget assertion that fails only inside the full parallel run on a loaded machine. Verified as pre-existing flake, not caused by this change: **3/3 passes in isolation on this branch and 3/3 on clean `main`**, and it passed in one full run and failed in another.
- Verified end to end by driving a real stdio `initialize` against the built `dist/` in two projects.

### DIFF
--- a/src/ai/embeddings.ts
+++ b/src/ai/embeddings.ts
@@ -3,6 +3,7 @@ import fsPromises from 'node:fs/promises';
 import { ProjectConfig } from '../core/types.js';
 import { KnowledgeEmbedder } from '../store/vector-index.js';
 import { fingerprintProfile, resolveVectorProfile, type VectorPooling } from '../core/vector-profile.js';
+import { noteModelLoad } from '../core/startup-trace.js';
 
 type TransformersPipeline = (texts: string[], options: { pooling: VectorPooling; normalize: boolean }) => Promise<{
   data: Float32Array | number[];
@@ -124,6 +125,10 @@ export async function createLocalEmbeddingProvider(
       .then(() => true)
       .catch(() => false);
     options.onFirstLoad?.({ model: vector.model, cacheDir, cached });
+    // How long building the pipeline actually costs, recorded rather than assumed. The model
+    // was blamed for stalls it cannot cause -- serve never loads one during startup -- and the
+    // only way that claim gets settled is a number per process, per model, cold and warm.
+    const loadStartedAt = Date.now();
     if (options.loadPipeline) {
       localPipeline = await options.loadPipeline(vector.model, vector.dtype, cacheDir);
     } else {
@@ -133,6 +138,7 @@ export async function createLocalEmbeddingProvider(
         dtype: vector.dtype as any,
       }) as TransformersPipeline;
     }
+    noteModelLoad(vector.model, cached, Date.now() - loadStartedAt);
     localPipelineKey = pipelineKey;
   }
 
--- a/src/cli/startup-report.ts
+++ b/src/cli/startup-report.ts
@@ -1,0 +1,130 @@
+import fs from 'node:fs';
+import { startupLogPath } from '../core/startup-trace.js';
+
+/**
+ * Read the startup trace back as an answer, not as a log.
+ *
+ * The question this exists to settle is "why did the server drop, and did the change fix it" —
+ * so it reports the two things that decide that: how long boots take now, and which phase owns
+ * the time. A boot that started and never ended is the signature of a host kill, and it is
+ * counted separately, because a kill leaves no summary line of its own to count.
+ */
+
+interface Record_ { event: string; [k: string]: any }
+
+const percentile = (sorted: number[], p: number): number =>
+  sorted.length ? sorted[Math.min(sorted.length - 1, Math.floor((p / 100) * sorted.length))] : 0;
+
+export function formatStartupReport(sinceHours: number): string {
+  const file = startupLogPath();
+  if (!fs.existsSync(file)) {
+    return `No startup trace yet at ${file}\nIt is written by \`knowl serve\`; start a session and re-run.`;
+  }
+
+  const cutoff = Date.now() - sinceHours * 3600_000;
+  const records: Record_[] = [];
+  for (const line of fs.readFileSync(file, 'utf8').split('\n')) {
+    if (!line.trim()) continue;
+    try {
+      const r = JSON.parse(line);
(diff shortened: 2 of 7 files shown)


dat999zx:
Merged to `main` — thank you, this was a good catch and a well-argued one.

Landed as your two commits, authorship intact:
- `814b5a9` feat(diagnostics): record why a serve startup was slow
- `c26afdc` fix(serve): complete the handshake before opening the database

Closing the PR rather than merging the button, because two follow-up commits went on top and `main` now carries the combined result.

**What changed on top of yours:**

`7684c8e` — three of the process hooks in `startup-trace.ts` altered behaviour rather than recording it:
- Listening for `unhandledRejection` suppresses Node's default (crash since v15), so the trace was quietly converting a fatal bug anywhere in the process into a logged one. Now re-thrown after recording.
- `process.exit(0)` on a signal reported a host kill as success — the opposite of what this module exists to distinguish. Now removes its listener and re-raises, falling back to 128+n.
- The durable log had no ceiling: one record per boot, per session, forever, and `diagnose-startup` reads it back in full. Capped at 4 MB, trimmed oldest-first on a newline boundary.
- The banner read `projectRoot=pending` permanently, which makes every serve process in a host log anonymous. The resolved root now arrives on a second line with the time it took.

To make the first two testable at all, the hooks take an injectable target. Eight tests added — the PR had none for 491 lines and a new core module.

`84ff934` — the comment justifying the restructure gave `busy_timeout` as 10s with five retries in `acquireClient`. Those are your fork's numbers; this repo sets `busy_timeout = 5000` and has no retry loop. Your description was straight about that, but the comment that ships is what the next reader believes, so it now says which numbers are whose. The argument doesn't depend on them.

**On the follow-up you offered at the end** — gating the whole of `bootstrapSchema` behind a cheap check: #14 did exactly that, and it's also merged. Verified together: 1361 tests passing.

dat999zx:
Landed on main — see the comment above.

PR #14, opened 2026-08-03
fix(store): gate schema bootstrap on a version, and elect one migrator

williamttruong:
`bootstrapSchema` runs in full on every read-write open: `migrateLegacyProjectSchema`, ~40 `CREATE TABLE IF NOT EXISTS`, ten `PRAGMA table_info` + conditional `ALTER TABLE ADD COLUMN` passes, `backfillKnowledgeAssertions` and `repairSkillForeignKeys`. Knowl runs one process per connected session plus whatever is started from a terminal, so that is many processes doing the same setup at the same time.

**The defect is not cost. It is a check-then-act race across processes.**

## Measured first, because the obvious diagnosis was wrong

Bootstrap is not slow — **37ms / 54ms / 106ms** over three runs on a real 317 MB database. And almost none of it takes a lock. Probing each statement shape against a deliberately held `BEGIN IMMEDIATE` from a rival process:

| statement | vs a held write lock |
|---|---|
| no-op `CREATE TABLE IF NOT EXISTS` | 2ms — free |
| no-op `CREATE INDEX IF NOT EXISTS` | 0ms — free |
| `PRAGMA table_info` | 31ms — free |
| `PRAGMA journal_mode = WAL` when already WAL | 0ms — free |
| a real `CREATE INDEX` | **7,115ms — blocked** |
| `backfillKnowledgeAssertions` INSERT..SELECT inserting **0 rows** | **2,430ms → SQLITE_BUSY** |

SQLite compiles a no-op `IF NOT EXISTS` to a *read* transaction (`Transaction` opcode P2=0), so the ~40 CREATEs and the ten `table_info` reads cannot be contending at all.

## The actual bug

The ten column passes read `PRAGMA table_info` and then conditionally `ALTER TABLE ADD COLUMN`. Two processes that both observe a column missing both issue the ALTER, and the loser dies on **`duplicate column name` — a `SQLITE_ERROR`**, which no `busy_timeout` and no BUSY retry can perceive.

Reproduced against this code at **1 in 96 cold-start opens** (12 concurrent processes × 8 fresh databases). It can only fire when a column is genuinely absent — that is, on the first opens after a release adds one, which is exactly when many sessions restart together.

## The change

1. Read `PRAGMA user_version` first — an O(1) header read, lock-free in WAL — and return immediately when current. **The steady-state open path becomes read-only.**
2. A database that does need work takes `BEGIN IMMEDIATE` and **re-reads the version under the lock**, so processes queued behind the winner find the work done and leave.

Measured over eight racing processes: `IMMEDIATE` elects one migrator and seven skip cleanly; `DEFERRED` kills seven of eight (a read→write upgrade is answered with `SQLITE_BUSY_SNAPSHOT`, which the busy handler is not allowed to wait out, per the deadlock-avoidance rule in `busy_handler.html`); no election at all applies the migration eight times.

Everything commits as one transaction, stamp included, so a process killed mid-migration leaves the database exactly as it found it.

**`PRAGMA foreign_keys` is silently ignored inside a transaction** — verified, it still reads `1` after being set to `0` — so the table rebuilds in `migrateLegacyProjectSchema` and `repairSkillForeignKeys` cannot use it here. `defer_foreign_keys` is the in-transaction equivalent, and it has the better failure mode: an inconsistent rebuild fails at COMMIT rather than persisting.

## `KNOWL_SCHEMA_VERSION` goes to 2, and the rule inverts

The current doc says additive columns need no bump, which was correct reasoning about an older client *reading* a newer database. It stops being correct the moment this number gates whether migrations run at all: **an additive column that does not bump the version is a migration every existing database skips forever.** Same for data backfills, which no schema comparison can represent.

Existing databases are stamped 1, so they run the full pass exactly once and land on 2.

`tests/store/schema-pin.test.ts` enforces this instead of relying on memory — it hashes the schema a fresh bootstrap produces and fails, with instructions, if that hash moves without the version moving.

## Verification

- Full suite: **1344 passed, 3 failed** — two `upgrade-all` tests that exceed the 30s default (measured **slower on unpatched `main`**: 35.5s / 40.7s there vs 33.7s / 30.7s here) and one `transcripts/backfill` timing assertion that **also fails on unpatched `main`**. None introduced here.
- Upgrade rehearsed on a copy of a real 317 MB database: version 1 → 2, 465 items and 465 embeddings intact, `foreign_key_check` clean, `integrity_check ok`.
- A subsequent open of that database completes bootstrap in **33ms while a rival process holds the write lock**.
- Cold-start race after the change: 0 failures, and every resulting database verified complete and correctly stamped.

## One thing deliberately not fixed here

`releaseAll()`'s `PRAGMA wal_checkpoint(TRUNCATE)` needs the write lock, so **process exit** can still block on a concurrent writer — measured at 4,877ms against a held lock. That is pre-existing and separate from the open path; flagging it rather than folding an unrelated change into this PR.

### DIFF
--- a/src/store/bootstrap.ts
+++ b/src/store/bootstrap.ts
@@ -1,7 +1,29 @@
 import { Client } from '@libsql/client';
 import { DEFAULT_FRESHNESS, hashKnowledgeContent, normalizeAffectedPaths } from './freshness.js';
-import { assertSchemaSupported, stampSchemaVersion } from './schema-version.js';
+import { assertSchemaSupported, KNOWL_SCHEMA_VERSION, readSchemaVersion, stampSchemaVersion } from './schema-version.js';
 
+/**
+ * Is this database already at the version this build ships?
+ *
+ * `PRAGMA user_version` is a 4-byte read at a fixed offset in the file header, so this is
+ * O(1) regardless of database size, and in WAL it never blocks on a writer. A database
+ * *newer* than this build is not handled here -- `assertSchemaSupported` refuses it before
+ * this is reached, because "newer than me" must be a refusal rather than a shrug.
+ */
+async function isSchemaCurrent(client: Client): Promise<boolean> {
+  return (await readSchemaVersion(client)) === KNOWL_SCHEMA_VERSION;
+}
+
+/**
+ * Per-CONNECTION setup, and deliberately outside the version gate below.
+ *
+ * `busy_timeout` and `foreign_keys` are connection state, not file state: they are not
+ * persisted, so every connection has to set them however current the schema is. Re-issuing
+ * `journal_mode = WAL` on a file already in WAL is free -- measured lock-free against a held
+ * write lock -- and WAL *is* persisted, so this is the one statement here that usually does
+ * nothing. It stays because libsql does not default to WAL, and the version gate's cheap
+ * lock-free read depends on being in WAL.
+ */
 const BASE_STATEMENTS = [
   // Must come first: journal_mode = WAL (and everything after it) takes locks, and a
   // connection's default busy_timeout is 0. Applied last, a concurrent writer at that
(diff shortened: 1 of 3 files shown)


dat999zx:
Merged to `main` — the measurement-first framing was the right call, and it changed the diagnosis. "Bootstrap is not slow, it is a check-then-act race" is a much better bug than the one this looked like from outside.

Landed as your commit, authorship intact:
- `b886e3b` fix(store): gate schema bootstrap on a version, and elect one migrator

Closing rather than merging the button, because one follow-up went on top.

**What changed on top of yours** — `84c9f05`:

The version bump needed splitting in two. `KNOWL_SCHEMA_VERSION` going to 2 under the new "bump on every additive change" rule is a trap, because that same number is what `assertSchemaSupported` refuses on: it rejects any database stamped higher than the running build. So every future release that adds a column would lock out every Knowl already installed — over a column older builds would have ignored, which is exactly what the old rule existed to permit. The first session to open a repo makes it unopenable by every other one.

Your reasoning for inverting the rule is right; it just proves the number is doing two jobs. "Can an older build read this file" moves rarely. "Has my migration run here" moves constantly. One value cannot be honest about both — making it truthful for the writer makes it a lie to the reader.

So the gate got its own number, `KNOWL_MIGRATION_LEVEL` in `PRAGMA application_id` — the other 4-byte header field, the same O(1) lock-free read in WAL you relied on, and never read by older builds, which is precisely the property `user_version` can't have. `KNOWL_SCHEMA_VERSION` returns to 1 with its original rule. Level 0 means "written before the gate existed", so existing databases still migrate exactly once, as you intended.

Two details worth naming: stamping is upward-only, since writing a lower level over a newer one would make that build redo its whole suite on every open once two versions share a file; and `isMigrationCurrent` is `>=`, so a database a newer Knowl migrated counts as current.

`schema-pin.test.ts` keeps doing its job, keyed on the migration level instead. Rehearsed on a real 567-item store: `user_version` 1 → 1, `application_id` 0 → 1, items and embeddings intact, `integrity_check` ok, `foreign_key_check` clean, second open performs no DDL.

Merged together with #13: 1361 tests passing.

dat999zx:
Landed on main — see the comment above.

PR #15, opened 2026-08-04
Audit remediation: 86 findings, all closed — security, correctness, integrity and cost

williamttruong:
Seven parallel read-only audits over 25,331 lines produced 70 reproduced findings. Sixteen implementation lanes closed them, and a second round that re-tested the first round's *dismissals* found more.

**86 findings, none left open. 1,725 tests passing (from 1,351), 0 failures. `tsc --noEmit` 15 → 0.** 67 source files, +8,115 / −3,183.

Full ledger with every finding, its mechanism, its reproduction and its outcome: `docs/audit-2026-08-04.md`. Narrative summary: `docs/audit-2026-08-04-report.md`.

---

## Read this part first, if nothing else

The audit's first round closed with items recorded as *"not worth fixing"*, *"needs a decision someone else must take"*, *"not testable"* and *"cosmetic"*. **None of those verdicts had been tested.** A second round ran under one rule: a conclusion that something should not be fixed is a claim, and needs evidence exactly like a fix does.

It found worse than the round it was checking:

| recorded as settled | what testing it showed |
|---|---|
| "DEP0190 cannot fire — it is the rationale, not a detector" | Fires on every `knowl doctor`, from a file the audit never opened |
| "Ten pre-existing type errors, cosmetic" | `@libsql/client` **throws** on an `undefined` bound parameter — four were crash sites |
| "Known-flaky wall-clock tests, do not chase" | The best-evidenced defect in the audit (K-65) |
| "`reindex-scope` fails on machine load" | 10,050 WAL fsyncs — a genuine 116 s operation under a 120 s timeout |
| "The only lever is shrinking what is scanned" | A write-time join table: **2100×** |
| "A running `serve` is protected by the OS" | False on Windows — `unlink` on an open file succeeds |
| "Not testable without a real model on disk" | Testable in ~2 s — and it hid a **live** defect |
| "The remedy is to stop moving 30 MB of BLOBs" | Transfer is ~25 ms of it; row-walking is the cost |

One of those was this codebase's own documented table, which was backwards in a way that would have steered every future batching decision wrong.

---

## Security and privacy

- **Importing a foreign export published that stranger's rows to your peers** — no join, no promote, no flag. Peers filter a repo's database on `visibility` alone, with no origin predicate. Export format → v3, carrying the exporter's identity; a foreign file may now update content but never ownership or visibility.
- **`.cmd`/`.bat` skill entrypoints were injectable** (BatBadBut / CVE-2024-24576), and `.cmd` was the *promoted* path — the tests used it as the canonical example. Dropped rather than escaped: a hand-rolled escape was attempted during research and was still injectable on the first try. `quoteShellArg` deleted; `autoRun` now defaults false.
- **Reading a peer created a database inside that peer's repo** (`file:` creates; `query_only` applies only after open).
- **The ownership guard failed open** when the owning peer was not checked out — and the obvious fix still missed the ledger's own example, because a manifest entry with no local path never becomes a peer at all.
- **`$HOME` was treated as a knowl repo**, because `~/.knowl` is indistinguishable from a project marker. Reproduced live during remediation, into a real home directory.
- **The MCP SDK validates the call envelope, never a tool's own `inputSchema`** — every constraint eleven tools had declared was decorative. One validator at dispatch closed seven findings.

## Correctness

- **The relevance floor deleted answers.** Its threshold had never been swept: **23 of 110 answerable queries returned nothing**, Recall@10 0.9818 → 0.7909. It now *reports* instead of deleting, so recall and MRR are identical at every threshold and the constant no longer has to be right.
- **Additive boosts on a fused rank score are structurally broken, not mistuned** — RRF destroys magnitude, so there is nothing for an additive constant to be a proportion of. Replaced with a convex combination (alpha swept over 2,149 cases), floor moved onto the raw cosine before fusion, priors made bounded multipliers.
- **Batching changed the stored vector**, so write-time embedding and `reindex --vectors` disagreed. It *did* change answers: over 120 real queries, top-10 order moved for 13.
- **Transcript search returned zero hits** for any query containing `.`, `-` or `/`; **discovery missed 282 files** (34.8%).
- **`doctor` could not see the gap it was written to find** — its coverage check omitted the `profile_fingerprint` predicate every read path applies.

## Data integrity

Snapshot restore cleared 5 tables while the delete cascaded into 8, leaving restored items permanently uneditable while reporting success. A failed batch ingest left rows with no commit record. GC hard-deleted the *richer* of two duplicates. `withClientTransaction` raised `SQLITE_ERROR`, not `SQLITE_BUSY`, so no retry recognised the collision. The `opening` transcript migration stranded any index predating it, permanently.

## Cost

| | before | after |
|---|---|---|
| Embedding reindex (10,050 rows) | 88,549 ms | **2,220 ms** |
| Blast-radius scan @ 100k commits | 14.2 ms | **0.007 ms** |
| Transcript hit rendering | 218 ms / 15.8 MB | **3.8 ms / ≤64 KB** |
| `commandExists` (6 lookups) | 1,483 ms | **38 ms** |
| Model cache on disk | 3,187 MB | 335 MB after horizon |

---

## Method

Lanes were partitioned by **file ownership**, not finding count — ten findings in one file is a merge disaster however independent the bugs are. Every lane wrote the failing test first; a fix was not done until reverting it turned the test red. Each lane was reviewed against the mechanism the ledger names, then merged one at a time with a full suite between.

**Four tests were found asserting the defects they covered**, including the secret-leak regression guard, which sets no environment variable — so `${OPENAI_API_KEY}` resolves to `''` and it passed whether or not the fix was present. A green suite is not evidence on its own.

**Things deliberately not done, each with the measurement**: ANN indexing (38× database growth to query *slower* at 97.5% recall), an in-memory vector cache (a 2.3× loss once cross-process invalidation is honoured), fp32 embeddings (3× per query for what `maxBatch: 1` gives free), and a demote path (mechanically one column; what is irreversible is semantic). K-56 did not reproduce, and the attempt is what corrected the transaction table.

## CI

Verified by running the workflow as it runs — **ubuntu, node 22** — not by trusting a green Windows suite. That found two tests that encoded the platform they were written on: a snapshot-path regex matching only a drive letter, and a 1 ms budget racing `deadline > startedAt`. Both fixed; two consecutive full ubuntu runs, 207 files, 1,723 passed, 2 skipped (Windows-only cases), exit 0.

---

## Notes for review

**This branch is 9 commits behind `main` and will show conflicts.** They are not mechanical, which is why they are not resolved here:

- `b886e3b` / `84c9f05` gate schema bootstrap on a version and elect one migrator, while this branch bumps the schema version for a new `knowledge_commit_items` table. Two independent designs for the same concern.
- `06bbf16` (`reindex only what a profile has no current vector for`) overlaps this branch's vector work, which moved cosine into SQL and batched the reindex writes.

Both want a deliberate reconciliation by someone who knows the intent behind the upstream changes. Resolving them blind is exactly the unverified change this whole audit was about avoiding — so they are flagged rather than guessed at.

**Still open**, recorded rather than hidden: `src/transcripts/embed-pass.ts` shares K-71's non-reproducibility (its batch composition varies with the catch-up deadline); the atom path took `maxBatch: 1` for free, but the transcript path is short-message shaped, where unbatching costs a measured ~2.7×. That needs a decision, not a patch.


williamttruong:
CI evidence, since workflow runs on fork PRs need maintainer approval:

**Green on real GitHub Actions** — the same `.github/workflows/ci.yml` (ubuntu-latest, node 22, `npm ci` → `npm run build` → `npm test`), dispatched on the fork:
https://github.com/William-Sommers/knowl/actions/runs/30894173327 — **success**

Also verified locally in the same environment (ubuntu 24.04, node 22.23.2) before pushing, twice consecutively: 207 files, 1,723 passed, 2 skipped, exit 0. The 2 skips are Windows-only cases (`it.runIf`).

Running CI as CI actually runs it — rather than trusting a green Windows suite — is what caught the last two defects in this branch. Both were tests that had encoded the platform they were written on:

- `restore-failure` matched the snapshot path with `/([A-Za-z]:[\/]...)/`, a drive letter. On Linux nothing matched and the assertion failed for the one reason it was not written to catch.
- `index-pass` passed `deadline: Date.now() + 1`. The guarantee under test is `budgeted = deadline > startedAt`, so a 1 ms budget is a coin flip: when more than a millisecond elapses before the pass records its own start — as it does on a loaded runner — the pass correctly reads a real budget as one already spent and indexes nothing.

Neither was a source defect. Both are fixed in `24bb6d7`.

dat999zx:
Landed on `main` as of 7f51976. Closing this rather than button-merging, because the reconciliation had to happen on a branch this repo controls — the same route #13 and #14 took. **All 100 commits are on `main` with your authorship intact**; nothing was squashed or rewritten.

Thank you for this. The ledger, the reproductions, and especially the second round that re-tested the first round's *dismissals* are the reason it could be reviewed at all rather than taken on trust. Verified independently before merging: build clean, `tsc --noEmit` 0 errors, 207/207 files and 1725/1725 tests on Windows — every claim in the description checked out.

## The two conflicts you flagged, and how they went

You said resolving them blind was the one thing to avoid. Both are resolved with reasons rather than guesses.

**Schema version — `main`'s design kept, and this one reverted.** We reached the same diagnosis independently: one number cannot mean "up to date" for the reader and "nothing to do" for the writer. You collapsed them and bumped `KNOWL_SCHEMA_VERSION` 1 → 3; `main` had split them into a compatibility floor (`user_version`) and a migration gate (`application_id`). The split is what shipped, because the bump would have been a regression: `assertSchemaSupported` refuses any database stamped above the running build, so raising `user_version` for an added table locks every 2.16 and 2.17 install out of its own data over a table they would have ignored. `knowledge_commit_items` is additive, so it moved `KNOWL_MIGRATION_LEVEL` to 2 and left the floor at 1. Your `backfillCommitItems`, the `busy_timeout` 5000 → 10000 correction, and the new table all landed unchanged.

**Vector reindex — both halves kept, plus a fix for their interaction.** `06bbf16` on `main` skips any item whose stored row already carries the current profile fingerprint; your work makes the reindex embed one text per pass so it reproduces write-time embedding exactly. Individually correct, and **jointly wrong**: a row written by a build that batched carries the same fingerprint and a materially different vector, so the skip would have preserved exactly the disagreement `maxBatch: 1` exists to remove — permanently, silently, and only on stores that upgraded rather than started fresh. Fixed by making the batch policy part of the vector's identity: `EMBEDDING_BATCH_POLICY` joins `provider|model|dtype|pooling` in `fingerprintProfile`, so rows written under the old policy invalidate the way a model change invalidates them.

## Elsewhere in the merge

- **`startup-trace.ts`** → `main`'s, which is a superset: log trimming, an injectable `ProcessHooks` seam, an `unhandledRejection` that re-throws instead of swallowing Node's crash, and a signal handler that re-raises rather than exiting 0.
- **`mcp/server.ts`** → **yours**, which fixes a live defect `main` still had. `instructions: mcpServerInstructions(config)` read the *positional* argument, and production passes `null` there and the real config through `deferred` — a handshake card no client had ever seen. Settling root and config before construction is what makes it reachable, and it also lets the first stderr banner name its repository instead of printing `unresolved` and correcting itself a moment later.
- The serve doc comment moved with it. Its note that the 10s `busy_timeout` and retrying `acquireClient` belonged to "a fork" stopped being true the moment this merged — they are this repo's now, so the arithmetic it describes is ours.
- `main`'s three additions to `src/index.ts` (`reindex --force`, the "skipped … already up to date" line, `diagnose-startup`) were ported into `src/cli/program.ts`, where you moved the command surface.

## Two follow-ups on top

**Default preset back to `granite-small-en-r2`.** Your arctic measurement stands and is why arctic is in the table — but 305 MB against 52 MB, and 768 dimensions against 384, is a lot to hand someone who never chose it. Arctic is one `preset` key away. Existing repos were never affected either way, since `preset` lives in `NEW_PROJECT_CONFIG` rather than `DEFAULT_CONFIG`.

**The relevance floor is now per model, and `MIN_VECTOR_RELEVANCE` is gone.** Your floor-sweep was right that a fixed absolute cosine does not transfer, and right to make it report instead of delete — that decision is what made the rest survivable. Measuring all five presets on one corpus showed the constant was wrong on *every* one of them:

| preset | on-topic min | off-topic max | at the old 0.30 |
| --- | --- | --- | --- |
| arctic-embed-m-v2 | 0.1638 | 0.2275 | 24/110 real answers mislabelled |
| granite-small-en-r2 | 0.7637 | 0.7644 | never fired |
| granite-97m-multilingual | 0.7443 | 0.7552 | never fired |
| bge-small-en | 0.5399 | 0.5754 | never fired |
| minilm-l6-en | 0.2003 | 0.2392 | 10/110 mislabelled |

Granite's whole scale sits ~0.5 above arctic's, so on three presets — including the default — the `NO CONFIDENT MATCH` notice was unreachable and the feature was dead. Per model, all five now mislabel 0 of 110 while catching 11–14 of 15. The scale-free alternative (top's margin/ratio/z against the rest of the candidates) was measured and is worse on both models — a query that finds nothing still produces a peaked distribution. An unmeasured model gets `null` and does not abstain, rather than borrowing another model's number. `docs/evals/per-model-floor.md`.

## Still open, as you recorded

`src/transcripts/embed-pass.ts` shares K-71's non-reproducibility, and the transcript path is short-message shaped where unbatching costs a measured ~2.7×. Left as a decision, not patched.

Final state on `main`: 210/210 test files, 1752/1752 tests, `tsc` clean, and recall byte-identical before and after the floor change on both suites.

williamttruong:
## Round two: an improvement pass, and what it found anyway

The audit rounds hunted defects. This round asked what could be **better** — embedding model, retrieval technique, storage engine, suite quality, and the agent-facing surface — routed through registry enumeration rather than search, with a transfer check against this corpus's actual shape (~500 dense hand-written atoms queried by a coding agent, not web-scale document RAG).

**Ledger is now 99 findings, none open. Suite 1,725 → 1,799. `tsc --noEmit` 0.**
**CI green on real GitHub Actions** (ubuntu, node 22): https://github.com/William-Sommers/knowl/actions/runs/30912148715

### The results that change behaviour

**The response was underfed, and that — not retrieval quality — was the problem.** `MAX_ITEM_CONTENT_CHARS = 600` with the *default empty* truncation marker: 84–94% of items exceed it, so agents saw **27–46% of stored content, cut mid-sentence, unflagged**, while doctrine told them to answer from memory instead of reading files. And `affectedPaths` never left the compact allowlist — a test asserted its absence — though ~50% of items carry it. Measured against 1,758 real transcripts (61,983 tool calls, 2,082 knowl calls, 296 sessions): of 356 cases where an agent queried then opened a file within 3 calls, the result had named that file **17.1%** of the time. The 53.5% "reads files anyway" rate was the correct response to that payload.

**MMR was billing relevance as redundancy.** On a topical query the second-best answer *necessarily* shares vocabulary with the best. Past rank one a lexical score is ~0.08 of the corpus best while token-Jaccard between two notes on one subject is 0.3–0.5 — the terms were never commensurate. Replacing it with a 0.9 near-duplicate demotion: R@10 **0.845 → 0.964** (semantic), 0.979 → 0.994, 0.985 → 0.996; confirmed on the data-disjoint `benchmark:accuracy` (Recall@5 0.710 → 0.748).

**The "2–6 keywords" rule was wrong and binding.** Agents obey it (73.4% of 921 real queries). Ground-truth ablation: truncating to 6 words cost −4.7pp and −7.2pp hit@1; on-subject extension reached 100%. Count was the wrong variable. It was stated in eight places.

**`synchronous` was never set** — every database ran at FULL by inheritance. NORMAL: un-batched writes **4.19×**, 6-process contention p95 **6.161 → 0.198 ms**. NORMAL matched OFF within noise, so the fsync was the whole gap.

**The embedding reindex committed once per row** under WAL+FULL — 11.57 vs 0.088 ms/row, **131.9×**. That was also the true cause of a test three lanes had called load-flaky: 10,050 fsyncs is a genuine ~116 s operation under a 120 s timeout.

### Decisions recorded with numbers, not opinions

Keep `arctic-embed-m-v2 / q8` — from an enumeration of **1,545 ONNX models in both sort orders**, 69 loadable, 13 measured; challengers led by under two cases on a 110-case instrument at 7.2–14× query cost. The finding that decided it: `MIN_VECTOR_RELEVANCE = 0.30` is an **absolute cosine calibrated to this embedder** — both challengers put their entire off-topic band above it and would answer 10/10 deliberately off-topic queries with every test green. A model swap silently disarms abstention.

**Cross-encoder reranking measured negative** on this corpus — 30 variants, 7 pool depths, 14 fusion forms, all at or below control. **ANN stays rejected with a named crossover: ~25,000 vectors** (real stores ~500; Turso's own CTO shipped SIMD brute force *instead of* ANN). **fp32 declined**: removes the batch divergence exactly, but 3× per query for what `maxBatch: 1` gives free.

### The suite was not evidence

Mutation testing found what four hand-found tests were only the visible part of. **The K-02 secret-scanning fix itself had a comment and no test** — deleting `title`, `tags`, `alternatives` or skill `steps` from the scanned set survived every test. Also surviving: `normalized.length > 0 → >= 0` (one empty `secretPatterns` entry then refuses **every** write), `>` → `>=` on both length ceilings, emptying `PROTECTED_CATEGORIES` in `gc.ts`. 41 tests added, every listed mutant re-verified killed. Whole-tree mutation is 68 h and not viable; `--since` per commit is 15–35 min and would have caught all four.

### Four for four

Every time this round ran the real CI environment instead of trusting a green Windows suite, it found something: two tests that had encoded the platform they were written on, then **K-98** (the byte-offset backfill checked its deadline before touching the first file, so the archives large enough to need it were exactly the ones it never reached) and **K-99** (a strict `>` against a millisecond-granular boundary — one bug that read as flakiness because three different tests in one file failed at random depending on which side of a tick each landed).

The merge conflicts noted in the description still stand and still want a deliberate reconciliation rather than a guess.

williamttruong:
Correction to my previous comment: it ended by saying the merge conflicts "still stand", which was already false when I posted it — this PR was merged at 10:41Z and v3.0.0 is cut. I was working from a stale local view and did not re-check the PR state before commenting. Apologies for the noise.

The round-two work described above is **not** in that merge; it is 20 commits that landed after it. I will open a separate PR for it rather than leaving it attached to a closed one.

Also worth noting, since it bears on one of the findings: `7f51976 fix(retrieval): one relevance floor per model, because a cosine scale is one` addresses exactly what round two measured independently — `MIN_VECTOR_RELEVANCE = 0.30` is an absolute cosine calibrated to one embedder, and both challenger models put their entire off-topic band above it, which would have silently disarmed abstention while every test stayed green. Good to see it fixed at the root rather than as a constant.

PR #16, opened 2026-08-04
Improvement round: retrieval, engine, model and suite quality — measured, with the negatives

williamttruong:
Follow-up to #15. That round hunted defects; this one asked what could be **better** — embedding model, retrieval technique, storage engine, suite quality, and the agent-facing surface. It still found defects, because measuring anything carefully does.

**Rebased onto v3.0.0 — 0 commits behind, no conflicts.** Suite **1,826 passing**, `tsc --noEmit` **0**. CI green on real Actions (ubuntu, node 22): https://github.com/William-Sommers/knowl/actions/runs/30913074298

Ledger: `docs/audit-2026-08-04.md` (99 findings, none open). Round-one narrative: `docs/audit-2026-08-04-report.md`. New evidence docs under `docs/evals/`.

---

## The finding that reframes the rest

**Retrieval was not the problem. The response was underfed.**

`MAX_ITEM_CONTENT_CHARS = 600` with the *default empty* truncation marker: 84–94% of active items exceed it, so agents received **27–46% of stored content, severed mid-sentence, with no flag** — while doctrine told them to answer from Knowl rather than read files. And `affectedPaths` never left the compact allowlist (a test asserted its absence) though ~50% of items carry it.

Measured against this machine's own archive — 1,758 transcripts, 61,983 tool calls, **2,082 knowl calls across 296 sessions**:

- Median first `knowl_query` is the **2nd tool call** of a session; 0.2% return empty. The agent *does* ask.
- Of 356 cases where an agent queried then opened a file within 3 calls, the result had named that file **17.1%** of the time. It could not have been higher.
- So the 53.5% "reads files after querying" rate was the correct response to the payload, not disobedience.

The **"2–6 keywords" rule was also wrong and binding** (agents obey it 73.4% of the time). Ground-truth ablation on the shipped suites: truncating over-length queries to 6 words cost **−4.7pp** and **−7.2pp** hit@1; off-subject padding cost −24 to −37pp; on-subject extension reached 100%. Count was the wrong variable. It was stated in eight places.

A companion change makes withheld scores legible: `uncalibrated: 'lexical-only' | 'not embedded'`, because only **17 of 924** archived results carried a score and silence could not be told from "no opinion". 30 bytes/result, absent when a real score ships.

---

## Retrieval: nothing beat control, except where nobody had looked

Thirty cross-encoder variants, seven candidate-pool depths, fourteen fusion forms — **all at or below control**. Gold is already in the candidate set for **99.94%** of 1,649 cases, so deeper pools buy 1.6 cases for 5.2× latency.

The win was in the half nobody had measured. A cross-cut of **26 systems read in source** (Letta, mem0, graphiti, cognee, khoj, basic-memory, smart-connections, Continue, Cline, Zed, Weaviate, txtai, LlamaIndex, Haystack, LanceDB, Qdrant, Typesense, Vespa, vectra, Orama…) found knowl matched the field on all four shared traits and was an outlier on exactly one — diversity. That outlier was the bug.

**MMR on the lexical fallback was billing relevance as redundancy.** On a topical query the second-best answer *necessarily* shares vocabulary with the best. Past rank one a lexical score is ~0.08 of the corpus best while token-Jaccard between two notes on one subject is 0.3–0.5 — the terms were never commensurate, so raising λ shrank the damage without fixing the sign.

| suite (lexical path) | R@10 before | after |
|---|---|---|
| semantic-suite | 0.84545 | **0.96364** |
| retrieval-suite | 0.97933 | **0.99400** |
| retrieval-suite-v2 | 0.98483 | **0.99575** |

Confirmed on a data-disjoint instrument: `benchmark:accuracy` Recall@5 0.710 → **0.748**, nDCG@5 0.622 → **0.654**. Not a museum path — CLI, MCP and the context composer all degrade to it, and the composer refuses to fetch a model not already on disk, so every fresh-repo bootstrap pack was built this way.

Also fixed: **`benchmarks/memoryagentbench/runner.ts` had been dead** since the embedding-profile guard landed — it built its vector option without `profileFingerprint`, so every run died in the driver. That is why the newest recorded result was from 2.8.0.

---

## Engine: `synchronous` was never a decision

knowl set `busy_timeout`, `foreign_keys` and `journal_mode` on every connection and **never set `synchronous`** — every database ran at FULL by inheritance.

| shape | FULL | NORMAL | |
|---|---|---|---|
| un-batched write | 3.488 ms/row | **0.832** | **4.19×** |
| 6 processes, one file | 173 w/s, p95 6.161 ms | **337 w/s, p95 0.198 ms** | 0 BUSY either way |

NORMAL matched `synchronous=OFF` within noise, which identifies the fsync as the *whole* gap. The trade in user terms, from SQLite's own documentation: a crashed agent, a killed `serve`, Ctrl-C, a closed lid — **lose nothing**; only a power cut can drop the last seconds, and the file still opens clean. `mmap_size`/`cache_size`/`temp_store` showed no effect above noise and are left at defaults, with the null result recorded in a test so nobody re-runs it.

**A landmine we are safe from only by accident**, now pinned: `@libsql/client`'s `transaction()` hands its connection away and nulls it, so the next `execute()` opens a fresh one with SQLite's defaults — `busy_timeout` 10000 → **0**. Measured: a contended write that waits 1467 ms and succeeds instead fails in 42 ms with `SQLITE_BUSY`. `withClientTransaction` uses raw `BEGIN`/`COMMIT` for an unrelated drizzle reason; a future "simplification" to the official API would disarm `busy_timeout` process-wide.

**The embedding reindex committed once per row** under WAL+FULL — 11.57 vs 0.088 ms/row, **131.9×**. That was also the true cause of `reindex-scope > pages past the old 10,000 ceiling`, which three lanes had independently called load-flaky: 10,050 fsyncs is a genuine ~116 s operation under a 120 s timeout. **88,549 ms → 2,220 ms.**

---

## Decisions recorded with numbers, including the ones to *not* do

- **ANN stays rejected, with a named crossover: ~25,000 vectors.** Exact scan is linear at 20.4 ms/1,000 at 768-dim; real stores are ~500 atoms. Measured ANN at 5,000×768: 179.5 s to build, database **21 MB → 810 MB (38×)**, and queries *slower* than the exact scan (27.0 vs 22.1 ms) at 97.5% recall. Turso's CTO closed their own ANN issue with "fast exact search is what many use cases actually want", shipping SIMD brute force instead.
- **Cosine moved into SQL** (`vector_distance_cos`, float32 layout byte-identical to `encodeVector`, no migration): 10,000 rows 84.4 → 57.4 ms, top-20 identical at every size. The prior diagnosis was wrong and is corrected in the ledger — decomposition shows walking the rows is the cost and BLOB transfer is ~25 ms of it, not 78%.
- **In-memory vector cache declined**: 7–12× in isolation, but the only correct cross-process invalidation is `PRAGMA data_version`, which fires on any write by any process — a wash at real size and a **2.3× loss** at 10,000 rows.
- **fp32 declined**: removes the batch divergence exactly (2.22e-16), but costs 3.0× per query, 3.9× on weights, and invalidates every stored vector — for the same exactness `maxBatch: 1` gives free.
- **Embedding model**: enumerated **1,545 ONNX models in both sort orders** → 69 loadable at q8 → 13 measured. Challengers led by under two cases on a 110-case instrument at 7.2–14× query cost. Note the two legs came back **fully disjoint**, and the recency leg found no new base model at all — the honest opposite of the failure it exists to prevent, and only knowable because it was run.

**On the default-model change in v3.0.0**: this branch takes yours (`granite-small-en-r2`) — the merge kept it and the suite is green. Our measurement was arctic-as-incumbent on our corpus and did not include granite, so it is not evidence against your choice. What it *does* say is recorded as K-92: `MIN_VECTOR_RELEVANCE = 0.30` was an absolute cosine calibrated to one embedder (arctic's off-topic band tops out at 0.193, while two challengers put their *entire* off-topic band above 0.30 and would have answered 10/10 deliberately off-topic queries with every test green). Your `7f51976` fixes that at the root, which is the right shape.

---

## The suite was not evidence

Round one found four tests by hand that asserted the defect they covered. Mutation testing (Stryker 9.6.1, enumerated from 134 npm packages ranked by real download counts — npm's `score.final` is saturated at 1.0000 for 25 straight packages) found they were the visible part.

**The K-02 secret-scanning fix itself had a comment and no test**: deleting `title`, `tags`, `alternatives` or skill `steps` from the scanned field set survived every test. Also surviving:

- `normalized.length > 0` → `>= 0` — one empty `secretPatterns` entry then matches every value (`includes('')` is always true) and **refuses every write**.
- `>` → `>=` on both length ceilings — tested at 11 chars against a limit of 10, so a payload of exactly the documented size was never checked.
- Emptying `PROTECTED_CATEGORIES` in `gc.ts` — every decision, constraint and architecture note becomes purgeable, on the one GC action with no undo.
- 25 of 32 regex sibling drops (`gh[pousr]_`, PEM's optional `RSA`, `xox[baprs]-`).

41 tests added; every listed mutant re-run and killed under a control-gated probe (verdicts by exit code — the probe's own first version had the defect class it was hunting, reporting SURVIVED for everything because ANSI escapes made its regex unmatchable). Whole-tree mutation is ~49,000 mutants ≈ 68 h and is **not** viable; `--since` per commit is 15–35 min and would have caught all four hand-found tests. It also established that the **16 test files spawning `node dist/index.js` can never kill a mutant** — nothing rebuilds the bundle between them.

---

## Four for four on CI

Every time this round ran the real CI environment instead of trusting a green Windows suite, it found something:

1. A snapshot-path regex matching only a drive letter (`[A-Za-z]:[\\/]`).
2. A 1 ms budget racing `deadline > startedAt`.
3. **K-98** — the byte-offset backfill checked its deadline *before* touching the first file, and the index pass only invoked it when budget was left. Under load it filled nothing, pass after pass. The archives large enough to need it are exactly the ones whose passes run out of budget, so K-47's seek-by-offset fix could never reach them.
4. **K-99** — a strict `>` against a millisecond-granular `tier_since`. One bug that read as flakiness: three different tests in one file failed at random on ubuntu depending on which side of a tick each landed.

Both new tests pin semantics rather than timing, so they fail on every platform before the fix rather than only on a fast one.

---

## Still open, deliberately

`src/transcripts/embed-pass.ts` shares K-71's non-reproducibility — its batch composition varies with the catch-up deadline. The atom and transcript paths both took `maxBatch: 1` (measured **0.93×** on real messages: the 2.7× it was deferred on came from synthetic 45-token items, and real messages have a p99 of 10,005 chars that dominates and is already unbatched). One `import-commit-trail` failure seen once under load (`before=1`) shares no mechanism with K-97 and remains unexplained rather than filed as solved.


williamttruong:
CI evidence, since workflow runs on fork PRs need maintainer approval:

**Green on real GitHub Actions**, same `.github/workflows/ci.yml` (ubuntu-latest, node 22, `npm ci` → `npm run build` → `npm test`), on this exact head commit:
https://github.com/William-Sommers/knowl/actions/runs/30913074298 — **success**

Also run locally in the same environment (ubuntu 24.04, node 22.23.2) against a checkout with a real `.git`, since one test asks git about its own repository: **214 files, 1,797 passed, 2 skipped**, exit 0. The 2 skips are `it.runIf(WINDOWS)` cases.

This branch is **0 commits behind main** and merges cleanly — the v3.0.0 merge is included and the suite is green with it.

williamttruong:
Final update — the deep mutation sweep landed. **57 files, 1,839 tests passing, `tsc --noEmit` 0, CI green:** https://github.com/William-Sommers/knowl/actions/runs/30932629942

Full report: `docs/evals/mutation-deep.md`.

**`agent-query.ts` 299/299 mutants (whole file), score 55.2% → 64.9%. `knowledge-writer.ts` 246 mutants, lines 79–327 only, 43.9% → 54.9%.** Six added cases killed +29 ranker and +27 writer mutants — about 4× the mutants they targeted.

### The one worth reading first

**The confidence clamp carried a comment explaining exactly why it exists, and no test.** That is the third appearance of the same shape in this audit (K-02's secret-scanning fix and the K-10 secret-leak guard were the first two) — a fix documented in prose and never pinned. Worse here: `confidence-bounds.test.ts`, the file that looked like its coverage, asserts an out-of-range value *can* still be imported. The apparent coverage was documenting the hole.

Fourteen more confirmed, with line and mutation in the doc. A few that matter: near-duplicate overlap could silently become one-sided **containment** (a long note demoted off the page because a shorter one above it reused its words); the one-token-title exclusion `sameSubjectTitle` describes only in prose; a newly declared `conflictExclusive` answered "already held" and dropped; a skill whose held `steps` were never fetched, so an unchanged skill churned a version on every re-store while a genuinely revised one was dropped.

### One behaviour reported rather than fixed

**A write that names an item to retire loses to the dedup search.** When `supersedes` names X *and* the same-subject search finds Y, Y is retired and X stays active. It surfaced because a mutation test failed against **unmodified** code, which is what separated it from a test bug. I have not changed it: an explicit id is the caller stating intent and a same-subject match is the store inferring it, and which should win is a decision about your write contract rather than a defect to patch. K-103 in the ledger.

### Coverage, stated as a fraction on purpose

**~60% of `knowledge-writer.ts` is unmutated** — lines 1–78 and 328–626, which is where transactions, commit records and batch atomicity live. `tool-schema.ts` is at 0%. And `agent-query.ts` lines **407–447** (the vector branch of `selectCandidates`) have **no covering test at all** — 25 NoCoverage mutants, and the cheapest remaining win in the file.

Three method notes that generalise beyond this repo:

1. **A sweep survivor is not a finding.** Five of seven candidates died against the wider suite; they are listed so nobody re-investigates them.
2. **A static covering set is a lower bound.** `query-score-surface.test.ts` reaches `agent-query.ts` dynamically, is absent from the covering-set generator, and kills 2 of 3 `uncalibrated` mutants — one finding was retracted after whole-suite confirmation.
3. **The timeouts carry no information.** The DB-free pass timed out **0** times on the identical 299 mutants while the libSQL passes hit 16/31/35, and those lists include `StringLiteral → ""` mutants that cannot loop. Runner artifact; ranked below Survived.

Whole-tree mutation remains ~68 h and not viable. `stryker run --since` per commit is 15–35 min at this repo's median churn, needs `concurrency: 1` and must exclude the 16 `dist`-spawning test files, and it would have caught all four of the hand-found defective tests.

dat999zx:
Merged and shipped in [v3.1.0](https://github.com/dat999zx/knowl/releases/tag/v3.1.0). Thanks — the diversity finding in particular was the kind of thing a suite can stay green through forever.

Two notes for the record.

**CLA.** This PR was opened on 2026-08-04, a day before `CLA.md` and the CLA bot landed on `main` (`da301cd`). The bot fires on `pull_request_target`, and there has been no push since, so it never ran here and no signature is recorded against this PR. If you could reply with the sentence below, it puts it on record:

> I have read the Knowl Contributor License Agreement and I agree to it.

**One test needed rewriting.** `tests/transcripts/offset-backfill.test.ts` → *"a budget that was real at call time buys one batch"* failed about one full-suite run in five locally. `runIndexPass` computes `budgeted = deadline > startedAt` with `startedAt` as its own first statement, so `Date.now() + 1` only arms the guarantee when under a millisecond passes between the caller's clock read and the pass's — one descheduling gap turns it off, which is why it reads as random.

Widening the deadline can't fix it either. Measured: the budgeted pass there is the *second* one, the file is already indexed, and it reaches the backfill in **4 ms** — so the deadline has to be both live at `startedAt` and spent 4 ms later. There is no safe constant in a 1–4 ms window.

It now asserts the guarantee on `budgetWasReal` directly, with the `runIndexPass` wiring pinned separately using an already-expired deadline, and both directions verified to fail when the guarantee is removed. Same rule, no clock. Only the test changed — the fix itself was right.

Issue #17, opened 2026-08-04
Proposal: live change impact — detection now, the write gate separately


williamttruong:
Before I open a PR: I've built something sizeable and I'd rather check whether you want it in knowl at all than hand you a large diff to react to. It's split into three independent pieces so you can take any prefix of it — including none.

## What it does

Detects when one session's in-progress code change invalidates work another session is still holding, and — in the third piece only — refuses that session's next write to the affected file until it re-reads.

Not a notification. The notification version is measured at roughly zero effect: SWE-Touch found that announcing a conflicting edit was no better than staying silent, and *worse* for 3 of 4 models. The refusal is STORM's mechanism, at symbol granularity rather than file granularity, which is where STORM's own false positives come from — a session that read `createSession` isn't blocked from editing `destroySession` beside it.

**Why here rather than as a separate tool.** The substrate is already yours: the symbol index, drift detection, evidence links, and hooks that are already in the session. A standalone tool rebuilds all of it for no gain.

## The three pieces

**1. `indexFile()` + one transaction per file** — 1 commit.

A per-file entry point for the symbol indexer, which previously only walked the whole repo, plus wrapping its writes in the transaction helper that already exists. That second part is a live perf fix independent of everything else here. No schema, no config, no new surface. This one stands on its own merits even if you want nothing else.

**2. Detection, advisory only** — 13 further commits.

`work_read_sets` + `impact_findings` (2 tables, 4 indexes, additive); `KNOWL_MIGRATION_LEVEL` 2 → 3 with `KNOWL_SCHEMA_VERSION` left at 1 so installed builds still open the file; a `toolName` field on normalized hook events, because reads and writes were previously indistinguishable downstream; read capture from `Read`/`NotebookRead` only; certain-tier detection; a `CODE IMPACT` stanza *inside* the existing mid-turn card rather than competing for the single-occupancy slot; and one MCP tool, `knowl_impact`.

Off by default (`impact.enabled`, deliberately **not** in `DEFAULT_CONFIG` so it isn't merged into every repo on the machine). **Nothing is ever blocked.** Every failure path is advisory.

**3. The write gate** — 1 further commit.

The `PreToolUse` refusal. This is the first thing in knowl that can stop a user's tool call, and it's separate precisely so that taking detection doesn't commit you to taking this.

## The part you should push back on

Two honest limits on piece 3, and they're why it's last rather than bundled:

[...]

PR #18, opened 2026-08-05
Live change impact: detect stale reads, and refuse the stale write

williamttruong:
Implements the proposal in #17. Opened as a draft so it can be read before it is a review request.

Detects when one session's in-progress code change invalidates work another session is still holding, and refuses that session's next write to the affected file until it re-reads.

Not a notification. The notification version is measured at roughly zero effect — SWE-Touch found announcing a conflicting edit no better than staying silent, and *worse* for 3 of 4 models. The refusal is STORM's mechanism at **symbol** granularity rather than file granularity, which is where STORM's own false positives come from: a session that read `createSession` is not blocked from editing `destroySession` beside it.

## Reviewing this in three parts

The commits are ordered so the three pieces of #17 can be read — or taken — independently. Each boundary is also a pushed branch, so any prefix can become its own PR if you would rather land them separately:

| part | branch | commits | what it is |
|---|---|---|---|
| 1 | `impact/1-index-file` | 1 | `indexFile()` + one transaction per file. No schema, no config, no new surface. |
| 2 | `impact/2-detection` | 17 | Schema, read-set, detection, card stanza, `knowl_impact`. **Advisory — refuses nothing.** |
| 3 | `impact/3-write-gate` | 18 | The `PreToolUse` refusal. |

Part 1 is a live perf fix that stands on its own merits regardless of what you decide about the rest: `indexCode` previously walked the whole repo with no per-file entry point, and wrote each symbol in its own implicit transaction.

## What it costs the repo

- **2 tables** (`work_read_sets`, `impact_findings`) + 4 indexes, additive. `KNOWL_MIGRATION_LEVEL` 2 → 3; `KNOWL_SCHEMA_VERSION` stays at 1 so installed builds still open the file. No FK to `memory_sessions` on purpose — a cascade at session GC would delete the evidence a finding was justified, which is the precision denominator.
- **One config flag**, `impact.enabled`, default off and deliberately **not** in `DEFAULT_CONFIG`, so it is not merged into every repo on the machine.
- **One MCP tool**, `knowl_impact`, registered only when the flag is on.
- **One new normalized hook event** (`tool-precheck` ← `PreToolUse`) and an optional `denyToolCall` on the host profile, implemented for Claude Code only — every other host fails closed to allowing the call.

## Precision, measured

The bar in the plan is ≥95% over ≥40 adjudicated findings, and the plan also records that nobody in this space publishes such a number. `tests/store/impact-precision.test.ts`:

| locator kind | precision | recall | tp | fp | fn |
|---|---|---|---|---|---|
| `symbol://` — the tier allowed to refuse a write | **100.0%** | 89.5% | 17 | 0 | 2 |
| `file://` | 100.0% | 100.0% | 2 | 0 | 0 |

46 scenarios. Labels are fixed before the run and never shown to the detector, so precision is arithmetic over labels rather than a judgement made after looking at the output. Both the ≥95% and the ≥40-sample bars are asserted in the suite, so the number cannot later be quoted off a sample too small to support it.

The benign half deliberately includes adversarial cases — a licence header added above the read symbol, two functions swapped in order, an import inserted at the top — each of which would fire if line position had leaked into the signature hash. None did.

**The suite earned itself on its first run.** Deleted and renamed symbols never fired: the candidate set was built only from symbols a file has *now*, so a symbol the change removed could never be looked up — while the change card and the refusal text both already carried a "gone" case nothing could reach. That is the strongest invalidation in the design and it was worth 28.6 points of recall. Fixed by seeding candidates with the pre-change symbols, guarded so absence counts as deletion only when the file demonstrably re-parsed; otherwise a file caught mid-edit, which yields no symbols at all, would report every symbol in it as deleted at once.

## Safety properties of the gate

- **Nothing is discarded.** A denial costs one tool call, reissued after a re-read. That is what separates it from optimistic concurrency, which aborted trajectories for 0.93× speedup at 1.83× tokens.
- **One shot per stale belief.** The refusal releases the read-set rows it named, so a retry is never blocked twice. An agent that re-reads with `cat` updates nothing, and a gate that can trap an agent is strictly worse than no gate.
- **Fails open, without exception** — flag off, no session, unknown host, unreadable config, broken store, failed release.
- **Exits 0 when it denies.** Claude Code reads the verdict from stdout only on exit 0; a non-zero exit is treated as a crash and the deny is silently discarded. `agent-hook.ts` now says so where the temptation to signal failure would be.

## Known limitations, stated rather than buried

- **`Bash` writes are ungated** — this is the largest remaining hole. STORM concedes the identical one, and it is filed upstream as anthropics/claude-code#29709. The mechanism that would close it is described in the plan's §10 and deliberately deferred: it has its own unmeasured precision, and anthropics/claude-code#79440 means shell aliases can rewrite a command after the hook approved it.
- **Two recall misses**, both the symbol extractor rather than the detector, each diagnosed by direct measurement: no symbol is emitted at all for `function*`, and `export` is stripped from signatures so un-exporting hashes identically. They stay counted as misses in the number above.
- **The precision figure is from constructed scenarios, not a field sample.** It shows the detector does not fire on the benign edit shapes a formatter, a linter or a neighbouring agent produce. It says nothing about how often those shapes occur in real sessions, which only live adjudication of `impact_findings` rows can.
- Two open host bugs still bound enforcement: anthropics/claude-code#77708 (deny not enforced in Desktop/Cowork) and #78527 (a 2.1.210 regression where a deny ends the turn, filed for `type: "prompt"` hooks — this one is `type: "command"`).

anthropics/claude-code#78970 — which #17 originally named as the limitation that mattered most — **does not reproduce** on 2.1.221. Every subagent tool call fired `PreToolUse` including `Edit`, for both `general-purpose` and `Explore`, with main-thread calls in the same runs as the control. Details in the #17 follow-up.

## Tests

Full suite green apart from four failures that are **pre-existing on `main`** and unrelated to this work — `agent-reminder`, `claude-continuation-reminder`, the `cli` guidance-refresh case, and `resume-points`. Verified by running the suite on a clean `main` checkout with none of these commits applied; #16 appears to fix them. `lexical-path` fails only under full-suite load and passes in isolation.

`npx tsc --noEmit` clean on all three branches.

### DIFF
--- a/docs/change-impact-plan.md
+++ b/docs/change-impact-plan.md
@@ -3,19 +3,17 @@
 Plan for handling the case where one actor's in-progress code change invalidates another
 agent's unfinished work. Written 2026-08-05.
 
-> **What is in this branch, and what is not.** This document describes the whole arc, including
-> the part that is *not* here. This branch is **detection only**: the read-set, certain-tier
-> findings, the card stanza and `knowl_impact`. It is off by default and it refuses nothing —
-> every path through it is advisory, and every failure path allows the turn to proceed.
+> **What is in this branch.** The **write gate** of §7.5 — the `PreToolUse` refusal — on top of
+> the detection layer. This is the first thing in knowl that can stop a user's tool call from
+> happening, so the two sections to read before the code are §7.5's four non-optional properties
+> and §10's coverage ceiling, which names four open bugs in the host's own hook implementation
+> by number.
 >
-> The **write gate** of §7.5 — the `PreToolUse` refusal that is the mechanism the evidence
-> actually supports — is deliberately proposed **separately**, so that accepting detection does
-> not commit anyone to accepting enforcement. Two reasons it is split rather than bundled: its
-> precision is not yet measured against §9's ≥95% bar, and §10 records four open bugs in the
-> host's own hook implementation that bound what a refusal can currently guarantee. Detection is
-> unaffected by all four.
->
-> Read §7.5 and §10 as design, not as shipped behaviour.
+> It is off by default (`impact.enabled`), fails open on every failure mode, and refuses at most
+> once per stale read so it can never trap an agent. §9.1 carries the precision measurement the
+> gate's bar is stated in terms of — **100% over 46 adjudicated scenarios**, against a ≥95% bar —
+> and §10 says plainly what that number does *not* establish, which is prevalence in real
+> sessions.
 
 **This is v2. It contradicts v1 (same path, earlier today) on its central design choice.** v1
 proposed *detect → notify the stale agent*. A full prior-art pass found that notification is
--- a/docs/change-impact-plan.md
+++ b/docs/change-impact-plan.md
@@ -566,7 +566,7 @@ Each rejected with a reason, not a shrug.
 |---|---|---|---|
 | **D-1** | Build here vs. standalone | **Here.** 60–70% of substrate exists [C]; niche unoccupied [W]; standalone means rebuilding it with no distribution. | provisional |
 | **D-2** | Upstream-shaped vs. fork-local | **Upstream-shaped from P-0** (§8 house rules), per the 2026-08-05 no-separate-version decision. | provisional |
-| **D-3** | Tell the upstream maintainer before or after P-3 | **Before.** Five tables in someone else's project; surprising a collaborator is worse than a slow start. | **William's — still open.** P-0…P-4 were built ahead of it, uncommitted; the ask is now larger, not smaller |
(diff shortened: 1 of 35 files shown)


williamttruong:
Closing this in favour of the three PRs you asked for — 1 shouldn't wait on review of 3, and you're right that they're independent decisions.

Piece 1 is up now as a separate PR. Pieces 2 and 3 are held until #16 lands, so their CI is readable rather than noisy, as you suggested.

Two things this PR's CI caught that are worth recording here rather than losing with the branch:

- **`snapshot-table-ownership` failed**, and correctly. `work_read_sets` and `impact_findings` weren't in `SNAPSHOT_TABLE_POLICY` — that file landed on `main` after I cut these branches, so my local suite never had the test. Both are now classified `preserved`, for the reason `memory_sessions` and `code_files` are: they describe live work against the working tree, and restoring them would resurrect a week-old session's beliefs about a file that has moved on. `impact_findings` has a second reason — its `resolution` column is the adjudication the precision number is computed from, so rolling it back would silently restate a measurement rather than merely lose one.
- **The CLA check is unsigned.** I'm not going to post the agreement text on William's behalf, so that's pending him personally.

All three branches are now rebased onto current `main` (they were 26 behind).

Replying to your review points on pieces 2 and 3 in #17, since they're design questions rather than diff comments.

PR #19, opened 2026-08-05
perf(code): indexFile(), and one transaction per file instead of per statement

williamttruong:
Piece 1 of the three in #17. Standalone: no schema, no config, no new surface, and defensible on its own merits whatever you decide about the rest.

Two changes to `symbol-index.ts`:

**A per-file entry point.** `indexCode` walked the whole repo and had no way to re-index one file. Anything wanting to react to a single write had to choose between a full walk and nothing.

**One transaction per file instead of one per statement.** The index wrote each symbol through a bare `execute`, which is its own implicit transaction — so indexing a file with 200 symbols paid 200 commits. This is a live cost independent of anything built on top of it, and it is why this piece is worth taking even if pieces 2 and 3 go nowhere.

`git checkout` between two branches with a few hundred changed symbols is the shape that shows it.

## Notes

- No behaviour change to `indexCode` itself; the walk calls into the same per-file path.
- `tests/code/index-file.test.ts` covers the per-file export, including that it respects `git check-ignore` the same way the full walk does, and that re-indexing a file replaces its symbols rather than accumulating them.
- Rebased onto current `main`.

### DIFF
--- a/src/code/symbol-index.ts
+++ b/src/code/symbol-index.ts
@@ -6,7 +6,7 @@ import { spawnSync } from 'node:child_process';
 import Parser from 'tree-sitter';
 import { CodeSymbol, CodeSymbolEdge, CodeSymbolKind } from '../core/types.js';
 import { CODE_EXTENSIONS, languageForExtension } from './languages.js';
-import { getClient } from '../store/database.js';
+import { getClient, withClientTransaction } from '../store/database.js';
 
 type SyntaxNode = Parser.SyntaxNode;
 type IndexedSymbol = CodeSymbol & { signatureHash: string | null };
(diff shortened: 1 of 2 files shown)


williamttruong:
The red `test` check here is a transient HuggingFace rate limit, not this diff.

The log carries `Error (429) occurred while trying to load file: "https://huggingface.co/Xenova/all-MiniLM-L6-v2/resolve/main/config.json"`, and the three suites that failed are the ones that need that model — `workspace-query`, `cross-repo-semantic`, `cross-repo-eval`. The `Unexpected token 'E', "Error exec"...` failures are downstream of the same fetch. Nothing in those paths touches `symbol-index.ts`, which is the only source file this PR changes.

I ran those three suites locally on this branch: 14/14 pass, model cached. I can't re-run the job from here (needs write on this repo), so it'll want a re-run or a later push to go green.

Worth knowing independently of this PR: CI depends on a live download from huggingface.co on every run, so any HF hiccup reds the build for reasons unrelated to the change under test. Caching the model in the workflow would make that class of failure go away — happy to open that separately if it's useful.

williamttruong:
I have read the Knowl Contributor License Agreement and I agree to it.

PR #20, opened 2026-08-05
ci: cache the embedding models so a HuggingFace hiccup stops failing the build

williamttruong:
Splitting this out of #19 since it isn't about that change.

CI loads a real embedding model and nothing caches it, so every run re-downloads it from huggingface.co. That puts an unrelated third party's rate limiter in the pass/fail path.

It bit #19, which changes exactly one file in `src/code/`:

```
Error (429) occurred while trying to load file:
"https://huggingface.co/Xenova/all-MiniLM-L6-v2/resolve/main/config.json"
```

`workspace-query`, `cross-repo-semantic` and `cross-repo-eval` went red together — the three suites that need the model, none of which that PR touches. All three pass locally against a warm cache.

## What this does

Caches `~/.knowl/models`, which is where `resolveModelCache` puts weights by default. Keyed on the two files that name the models (`vector-profile.ts`, `config.ts`), so switching or adding a model fetches once and then caches.

`restore-keys` is carrying more weight than the exact key: on a miss it still restores the newest previous cache, so a profile change re-downloads only what actually changed instead of the whole set.

## What it doesn't do

The first run after this merges is a cache miss by definition and can still hit a 429 — it's the runs after that this fixes. And it doesn't help if HF is down when the cache is cold.

If you'd rather pin this harder, the next step would be failing fast with a clear message when the model can't be fetched, so a network problem reads as a network problem instead of six assertion failures in suites that look unrelated. Happy to add that if you want it, but it's a bigger change to the embedding path and I didn't want to bundle it with a workflow tweak.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,27 @@ jobs:
           node-version: 22
           cache: npm
 
+      # The suite loads a real embedding model, so without this every run re-downloads it from
+      # huggingface.co and a rate limit there fails the build for reasons unrelated to the change
+      # under test. Observed on 2026-08-05: `Error (429) ... Xenova/all-MiniLM-L6-v2/config.json`
+      # took down `workspace-query`, `cross-repo-semantic` and `cross-repo-eval` on a PR that
+      # touched none of them.
+      #
+      # `~/.knowl/models` is where `resolveModelCache` puts weights by default -- shared per
+      # machine rather than per repo, because nothing about them varies by repository.
+      #
+      # Keyed on the files that name the models, so adding or switching one fetches once and then
+      # caches. `restore-keys` matters more than the exact key here: on a miss it still restores
+      # the newest previous cache, so a profile change re-downloads only what actually changed
+      # instead of everything.
+      - name: Cache embedding models
+        uses: actions/cache@v4
+        with:
+          path: ~/.knowl/models
+          key: knowl-models-${{ runner.os }}-${{ hashFiles('src/core/vector-profile.ts', 'src/core/config.ts') }}
+          restore-keys: |
+            knowl-models-${{ runner.os }}-
+
       - name: Install dependencies
         run: npm ci
 


dat999zx:
Thanks for splitting this out — the diagnosis is right, and more precisely right than it needed to be. `workspace-query`, `cross-repo-semantic` and `cross-repo-eval` are the entire set of suites that load a real model; everything else mocks `loadPipeline`, stubs the embedder, or pre-places fake weight files. Putting HF's rate limiter in the pass/fail path is a real problem and it's worth fixing.

The cache path doesn't work, though, and I want to be specific about why rather than just say no.

`~/.knowl/models` is indeed where `resolveModelCache` puts weights by default, but no test ever sees the real home:

- `vitest.config.ts` sets `KNOWL_HOME` to `./.knowl-test-home` for every worker. That's deliberate — suites that spawn the real CLI would otherwise write fixtures into a developer's home, where `knowl upgrade --all` would later go visit them.
- `resolveModelCache` tries `knowlHome()/models` before `<projectRoot>/.knowl/models`, so under test the first candidate is always inside the repo.
- `cross-repo-semantic` and `cross-repo-eval` then override `KNOWL_HOME` again to their own fixture (`./.knowl-semantic-home`, `./.knowl-xeval-home`), and `afterAll` deletes it. `.knowl-*/` is gitignored and swept by `tests/global-teardown.ts` besides.

So every directory the resolver could pick is gone before the job ends. The cache step would save nothing, restore nothing, and the 429 would come back — with the workflow now looking like it had been handled, which is the part that would have cost someone an afternoon later.

The missing piece is that there was no shared path to cache in the first place. `workspace-query.test.ts` and `write-embedding.test.ts` already point at the repo's own `.knowl/models`; the two cross-repo suites just never adopted it. #21 points them at the same cache and then caches `.knowl/models`, keeping your key design unchanged — `vector-profile.ts` + `config.ts`, with `restore-keys` doing the real work on a miss, for the reason you gave.

Verified cold in a clean worktree: the three suites pass and leave 23 MB of weights behind, so there's something to save. Warm full suite is green (226 files, 1906 tests).

Closing this in favour of #21. The finding is yours and the credit is too — it's noted in the PR body. This should also unblock #19.

On your offer to fail fast with a clear message when the model can't be fetched: yes, that's worth having, and you're right that it doesn't belong in a workflow tweak. Open it as its own PR if you'd like to.

PR #21, opened 2026-08-05
ci: cache the embedding model so a HuggingFace 429 stops failing the build

dat999zx:
Supersedes #20, which found this and got the diagnosis exactly right. Same problem, cache pointed somewhere the suite can actually reach.

## The problem

Three suites load a real embedding model and nothing caches it, so every CI run re-downloads it from huggingface.co. That puts an unrelated third party's rate limiter in the pass/fail path. On 2026-08-05 a 429 on `Xenova/all-MiniLM-L6-v2/config.json` took down `workspace-query`, `cross-repo-semantic` and `cross-repo-eval` together — on #19, which touches none of them.

Those three are the whole set. Everything else either mocks `loadPipeline`, stubs the embedder, pre-places fake weight files, or skips when the model is absent.

## Why caching alone doesn't fix it

The obvious path to cache is `~/.knowl/models`, where `resolveModelCache` puts weights by default. No test ever sees the real home:

- `vitest.config.ts` pins `KNOWL_HOME` to `./.knowl-test-home` for every worker — deliberately, so a spawned CLI cannot scribble in a developer's home directory.
- `resolveModelCache` (src/ai/embeddings.ts) tries `knowlHome()/models` before `<projectRoot>/.knowl/models`, so under test the first candidate is always inside the repo.
- `cross-repo-semantic` and `cross-repo-eval` override `KNOWL_HOME` again to their own fixture, which `afterAll` deletes. `.knowl-*/` is gitignored and swept by `tests/global-teardown.ts` besides.

Every directory the resolver would pick is gone before the job ends, so an `actions/cache` step on `~/.knowl/models` saves nothing and restores nothing.

## What this does

Gives the three suites one cache to share, and caches that.

`workspace-query.test.ts` and `write-embedding.test.ts` already point at the repo's own `.knowl/models`, with a recorded reason: the model loads once instead of once per fixture root. The two cross-repo suites now do the same — a `MODEL_CACHE` constant and a `CONFIG` that sets `search.vector.cacheDir`, used for `saveConfig` and every `createLocalEmbeddingProvider` call.

`.github/workflows/ci.yml` then caches `.knowl/models`. The key design is #20's and unchanged: keyed on `vector-profile.ts` and `config.ts`, the two files that name the models, so switching or adding one fetches once and then caches. `restore-keys` carries more weight than the exact key — on a miss it still restores the newest previous cache, so a profile change re-downloads only what actually changed.

## What it doesn't do

The first run after this merges is a cache miss by definition and can still hit a 429; it's the runs after that this fixes. It doesn't help if HF is down while the cache is cold.

## Verification

- Cold run in a clean worktree with no `.knowl`: the three suites pass 14/14 and leave 23 MB at `.knowl/models/Xenova/all-MiniLM-L6-v2` — so the cache step has something to save, and the restore has somewhere to land.
- Warm: full suite green, 226 files / 1906 tests. `npm run build` and `git diff --check` clean.

One deliberate side effect: with `.knowl/models` warm, `write-embedding.test.ts`'s `it.skipIf(!MODEL_CACHED)` case stops being skipped in CI. It passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,16 @@ jobs:
       # then caches. `restore-keys` matters more than the exact key here: on a miss it still
       # restores the newest previous cache, so a profile change re-downloads only what
       # actually changed instead of everything.
-      - name: Cache embedding models
-        uses: actions/cache@v4
+      #
+      # Split into restore and save rather than one `actions/cache` step, because that step
+      # only saves when the job succeeds. This suite has a known timing-sensitive test, so a
+      # save-on-green cache stays cold exactly through the runs that need it most: one flake
+      # discards the download, the next run fetches again, and the rate limit this exists to
+      # avoid is hit by the retry. The weights are valid whatever the tests concluded about
+      # the code.
+      - name: Restore embedding model cache
+        id: model-cache
+        uses: actions/cache/restore@v4
         with:
           path: .knowl/models
           key: knowl-models-${{ runner.os }}-${{ hashFiles('src/core/vector-profile.ts', 'src/core/config.ts') }}
@@ -62,3 +70,15 @@ jobs:
 
       - name: Test
         run: npm test
+
+      # `always()`, so a red suite still leaves the weights behind -- that is the whole point
+      # of the split above. Guarded twice: `cache-hit` skips the upload when the exact key was
+      # already restored, and `hashFiles` returns an empty string when nothing is on disk,
+      # which is what stops a job that died before the tests ran from failing again here on a
+      # missing path.
+      - name: Save embedding model cache
+        if: always() && steps.model-cache.outputs.cache-hit != 'true' && hashFiles('.knowl/models/**') != ''
+        uses: actions/cache/save@v4
+        with:
+          path: .knowl/models
+          key: ${{ steps.model-cache.outputs.cache-primary-key }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,32 @@ jobs:
           node-version: 22
           cache: npm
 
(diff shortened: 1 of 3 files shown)


dat999zx:
First CI run on this branch: red, but not from this change, and it made the case for a follow-up commit.

**What passed.** All three model suites — `cross-repo-eval` (4), `workspace-query` (6), `cross-repo-semantic` (4). The cache was a miss, as expected on a first run; the download succeeded and left 23 MB in `.knowl/models`. The mechanism works.

**What failed.** One test: `tests/transcripts/offset-backfill.test.ts:253`, `expected 0 to be greater than 0`. That is the known K-98 timing flake — `runIndexPass` captures `startedAt` inside the pass, so a deadline computed as `Date.now() + 1` in the test is already spent if a millisecond elapses first. Unrelated to caching, and being fixed separately.

**What that exposed.** `actions/cache` defaults to `save-always: false`, so the post step only saves on a green job. The flake went red and the 23 MB was discarded — the next run would have fetched it all over again.

That is the worst failure mode this cache could have. A save-on-green cache stays cold exactly through the runs that need it most: one flake throws the weights away, the next run refetches, and the rate limit this exists to avoid gets hit by the retry. With a known timing-sensitive test in the suite, that is not hypothetical — it happened on the first try.

`1db690c` splits the step into `actions/cache/restore` and `actions/cache/save` with `if: always()`. Whatever a red suite says about `offset-backfill`, it says nothing about whether all-MiniLM-L6-v2 downloaded correctly. Guarded twice: `cache-hit` skips the upload when the exact key was already restored, and `hashFiles('.knowl/models/**') != ''` stops a job that died before the tests ran from failing again on a missing path.

Workflow-only; the test-file half is untouched and still verified.

Issue #30, opened 2026-08-06
Global teardown sweeps the repo root but not os.tmpdir() — 39 mkdtemp sites leak, 5.7 GB / ~3 days on Windows


williamttruong:
**Corrected 2026-08-06** — my original report suggested "add a global teardown sweep." That sweep already exists and works. The real defect is narrower; the body below is rewritten to describe it accurately. Apologies for the noise.

---

`tests/global-teardown.ts` sweeps `process.cwd()` for **dot**-prefixed `.knowl-` entries. But **39 `mkdtemp` calls across 29 test files** create their fixtures in `os.tmpdir()` with a **dotless** `knowl-` prefix, which the sweep never visits. Only 4 call sites use the repo-root `.knowl-` convention the sweep knows about.

So two fixture conventions coexist and only one is collected.

### Measurement

On Windows, `%TEMP%` held **28,524 `knowl-*` directories totalling 5.7 GB** on 2026-08-06. Oldest dated 2026-08-03, newest the same day — roughly **1.9 GB per day for every day the suite is run**. Clearing them broke nothing, and they have not regrown since (the suite hasn't been run).

Largest groups by prefix, each traceable to a specific suite:

| prefix | dirs | source |
|---|---|---|
| `knowl-handlers` | 3,630 | `tests/transcripts/mcp-handlers.test.ts:55` |
| `knowl-pass` | 3,461 | `tests/transcripts/index-pass.test.ts:26` |
| `knowl-sem` | 3,350 | transcripts semantic suite |
| `knowl-lex` | 3,314 | transcripts lexical suite |
| `knowl-fed` | 2,258 | `tests/transcripts/federate.test.ts:14` |

### Why it is invisible on Linux and macOS

The per-test cleanup is best-effort by design, and correctly so. From `tests/transcripts/index-pass.test.ts:32`:

```js
afterEach(async () => {
  await closeTranscriptDbs();
  // Swallowed: Windows keeps the database locked for the life of the process. Each test has
  // its own mkdtemp root, so nothing leaks between tests.
  await fs.rm(dir, { recursive: true, force: true }).catch(() => {});
});
```

[...]

PR #33, opened 2026-08-06
feat(impact): detect when one session's change invalidates another's reads — detection only, off by default

williamttruong:
Closes the detection half of #17. This is **piece 2 of the three** you asked me to split it into — piece 1 landed as #19, and piece 3 (the write gate) stays parked until this one is decided, exactly as I said on #17.

Held until #16 landed so the CI signal here would be readable rather than noisy. #16 is in, so this is now cut from current `main` with a clean run.

## What it does

Detects when one session's in-progress code change invalidates work another session is still holding, and surfaces it. **Nothing is ever blocked** — every path in this PR is advisory. The refusal lives in piece 3 and is not here.

Off by default via `impact.enabled`, and deliberately **not** in `DEFAULT_CONFIG`, so it is not merged into every repository on the machine.

New surface: `work_read_sets` and `impact_findings` (2 tables, 4 indexes, additive), `KNOWL_MIGRATION_LEVEL` 2 → 3 with `KNOWL_SCHEMA_VERSION` left at 1 so installed builds still open the file, a `toolName` field on normalized hook events, read capture from `Read`/`NotebookRead` only, a `CODE IMPACT` stanza *inside* the existing mid-turn card rather than competing for the single-occupancy slot, and one MCP tool, `knowl_impact`.

## The three corrections from #17, now in the branch

These were posted as a comment on #17 and are worth repeating where the code is:

- **`anthropics/claude-code#78970` does not reproduce.** I originally cited it as the limitation that mattered most — that `PreToolUse` may not fire for subagent tool calls. I had taken that from the open issue rather than testing it. Testing it says the opposite: every subagent tool call fired `PreToolUse` carrying `agent_id` and `agent_type`, including `Edit`, on 2.1.221. The same run also disposes of `#79480` for this configuration. That correction makes the *gate* stronger, and it does not affect this PR either way.
- **`NotebookEdit` is fixed, not deferred.** It is the only write tool that does not name its target `file_path`, and `notebook_path` was missing from both the stdin allowlist and `changedPaths`. It did not fail, it went quiet.
- **Certain-tier precision is measured**, over 46 adjudicated scenarios with labels fixed before the run and never shown to the detector: `symbol://` 100.0% precision / 89.5% recall, `file://` 100.0% / 100.0%. That number was the main reason for proposing the gate separately, and it now exists.

## Rebase note

`main` moved 72 commits while this waited. One conflict, in `src/mcp/tools.ts`, all three hunks against my own #28:

- `TRANSCRIPT_TOOLS` → `TRANSCRIPT_TOOL_DEFINITIONS` — took the new name.
- `orderForSelection(tools)` — kept, and placed *after* the impact tools are pushed so ordering covers them too.
- `throw new Error('Unknown tool')` → `McpError(ErrorCode.InvalidParams)` — kept your stricter version, with the `knowl_impact` handler slotted into the `else if` chain above it.

I checked specifically that the resolution did not drop #28's fetch-by-id path; all three `getKnowledgeItem` call sites survive.

The final commit clears two lint warnings that were the branch's own, not rebase artefacts — a dead static import, and a `payload = {}` initializer that both the `try` and the `catch` overwrite (the same dead-default the audit removed elsewhere). They only became visible once this sat on a `main` that lints clean.

## Verification

- `npm run typecheck` — clean
- `npm run lint` — **0 errors**, and **0 warnings in any file this branch touches** (the 26 remaining are pre-existing on `main`, cross-referenced by file rather than by count or line number)
- `npm test` — **245 files, 2,141 passed, 4 skipped, 0 failed**

One transparency note on the test run: an earlier scoped run had `tests/mcp/workspace-query.test.ts` time out at 30s under parallel load. It passes in ~2s in isolation, which is exactly what a real bug also sounds like, so I re-ran the identical scope rather than accept the isolated pass — 261/261, including that test. The full-suite run above is green end to end.

## What I am asking

Whether you want detection. If yes and it merges, I will rebase piece 3 and open it separately so the gate stays its own decision. If not, it stays in my fork and #17 can close.

---

## Rebased again onto #31 (2026-08-06)

#31 (`refactor: make src/ a layered DAG, and enforce it`) landed while this branch's suite was running, so the first push was against the previous `main` and GitHub could not build a merge commit for it — which is why no CI ran on it at all. Re-rebased onto current `main`. Four conflicts this time, all from files moving rather than changing:

- `change-card` and `host-lifecycle` moved into `src/session/`, shifting every relative import here.
- The hook types moved to `core/host-hook-types.ts` behind a re-export, so this branch's `toolName` field moved there with them.
- `session-handoff`, `resume-points` and `resume-keys` moved to `src/session/`.
- `tests/store/impact-lifecycle.test.ts` imported two paths that no longer exist. Nothing conflicted — the imports simply resolved against moved files — so the rebase never surfaced it. Fixed and passing 10/10.

## One thing I did not decide for you

Your new `tests/architecture/module-boundaries.test.ts` caught exactly one violation in this branch, and it is right:

```
store/impact.ts imports '../code/symbol-index.js' (store -> code)
```

I **deferred** the import, which is one of the three fixes your error message names. It is defensible here rather than a dodge: this subsystem is off by default, and the import sits past the early return for "no changed paths", so the indexer loads only on a run that has something to compare — not on every process that opens the store.

The other fix is to move the module to its real layer, and I did not do that unilaterally because it changes *your* architecture declaration rather than my file. `impact.ts` depends on `store` and `code`, and both `session` and `mcp` consume it, so it cannot sit beside either — it needs a new layer between `code` and `session`:

```ts
['core'],
['store'],
['ai', 'workspace', 'skills', 'code'],
['impact'],                                    // <- new
['session', 'pipeline', 'transcripts', 'viewer'],
['cli', 'mcp'],
```

Say the word and I will move it and drop the deferred import. Only `impact.ts` crosses the line; `impact-config.ts`, `read-set.ts` and `working-tree.ts` import nothing from `code` and belong in `store` either way.

## Verification after the second rebase

- `npm run typecheck` — clean
- `tests/architecture` — **6/6**, gate green
- `npm test` — 245 of 246 files pass, **2,146 passed / 4 skipped / 1 failed**

The one failure is `tests/transcripts/teardown.test.ts`, and I do not believe it is this branch: the branch touches **zero** transcript files, that test file is byte-identical to `main`'s, it passes 8/8 in isolation, and it fails as `seed failed:` with an *empty* stderr — a spawned node process exiting non-zero saying nothing, which reads as local resource exhaustion rather than logic. A previous run failed `tests/mcp/workspace-query.test.ts` the same way and it has since passed three times, including 6/6 in isolation. I am flagging it rather than calling it clean, and deferring to what CI says on runners that are not this machine.

Related: this suite leaves its `mkdtemp` roots behind on Windows — I filed that separately as #30 (28,524 directories / 5.7 GB from about three days of runs), which is plausibly the same resource pressure.


williamttruong:
CI is green on all six checks, including `windows-latest / node 22` — the exact configuration where my local `tests/transcripts/teardown.test.ts` failure happened. So that was this machine, not the branch, as the empty stderr suggested. Nothing outstanding from my side.

The layering question above is still open and is the one thing I'd like a call on: deferred import as it stands, or move `impact.ts` into its own layer between `code` and `session`.

PR #34, opened 2026-08-06
fix(workspace): federated reads report their verdict; record cross-repo demand

williamttruong:
Two commits. The first is three read-side bug fixes that stand alone; the second adds a
measure-only ledger that depends on them being correct. Design and plan docs are included
(`docs/superpowers/specs/2026-08-07-demand-paged-scoping-design.md` and the paired plan).

## Why

A workspace makes knowledge reachable, and almost none of it is. Measured across three linked
repos: 634/31 shared, 223/18, and one repo at 100% via `defaultVisibility` — ~95% of the two
code repos' knowledge is unreachable from anywhere else. Sharing is a manual per-item act and
promotion is irreversible, so the default is private and the default is what happens. The
visible symptom is an asymmetric surface: seven rules exist word-for-word in both repos, and in
all seven the copy in one repo is shared while the other's is private. Nobody decided that.

"Promote more" is the wrong fix — it converts a reversible under-sharing problem into an
unfixable over-sharing one, and it declares scope statically before anyone knows what another
repo will ever ask for. This PR does not fix the scoping problem. It fixes the reads, and starts
recording the evidence that would justify doing anything further.

## 1 — `fix(workspace): a federated read now reports what it actually found`

**The relevance floor never reached federated ranking.** `rankKnowledge` passes
`vector.relevanceFloor` into `scoreCandidates`; `queryFederated` did not. So `minRelevance`
arrived `null` on every workspace query, `answerable` was unconditionally true, no federated
result could carry `abstained`, and `knowl_query`'s `NO CONFIDENT MATCH` notice was unreachable
code **from the moment a repo was linked**. That is the case where the verdict matters most,
because the alternative on offer is another repo's near-miss.

**`kin` was a write-time signal only.** `findCrossRepoOverlap` searches a kin peer wider and the
CLI advisory says "shares this repo's lineage" — so someone *storing* a fact was warned about
divergence and an agent *reading* one was not. Reading is where a diverged convention gets
applied to the wrong repo. Results from a kin peer now carry `kinDivergent`, and `knowl_query`
renders one `SHARED LINEAGE` block per response rather than one per row.

**The miss notice named no next move.** An abstention is the one moment the caller has concluded
memory is empty — exactly when transcript search can still answer, since sessions are indexed
separately and `knowl_query` never touches them. Now named, conditional on
`isTranscriptSearchEnabled`: pointing at a tool the build does not expose is worse than silence.

## 2 — `feat(workspace): record what the repos in a workspace ask each other for`

`src/workspace/demand-ledger.ts` — one SQLite file per workspace, beside the manifest rather
than inside any member repo, plus `knowl workspace demand` to read it. **Nothing acts on it.**

- **Every query is logged, not just weak ones.** The obvious design treats abstention as the
  "workspace could not answer" signal, but abstention is corpus-dependent: on a real 483-item
  store, off-topic queries top out around 0.29 against a 0.16 floor, so that predicate fires
  almost never and would yield an empty ledger read as "there is no demand". The score goes on
  every row; the threshold gets chosen from the distribution afterwards.
- **Privacy**: fingerprint always, query text only when it passes exactly the validators a
  knowledge write passes. This file lives outside every repo's ignore rules — it must not hold
  what the store next door would refuse.
- **Never fails a query**: writes are `void`-ed and swallow every error. A test drives the case
  by replacing the workspace directory with a file.
- **Low `busy_timeout` (250ms)**, deliberately unlike `resume-store`'s 10s — this is on the
  response path, and losing a row beats blocking a query.

One bug worth naming because a reviewer would not spot it: index creation must run **after** the
column guard, not beside the table creation. An index names columns, so on a file written by a
build that predates one, `CREATE INDEX` throws during open and the guard that would have
repaired it never runs. Caught by the legacy-file test, not by review.

## What is deliberately absent

The consolidator this ledger was meant to feed. Its entry gate has two halves; one now passes
and one does not, so it stays unbuilt.

**The duplicate matcher (passes, with a replacement).** The gate wanted `sameSubjectTitle` to
match at least 5 of 7 known duplicate pairs; measured, it matches 4. Containment is the wrong
shape — two independently-written titles for one rule almost always each carry a word the other
lacks, and one such word vetoes the match. So four candidates were measured against the seven
pairs *and* a hard-negative set, judged on separation rather than recall (`() => true` scores
7/7):

| matcher | worst positive | best negative | verdict |
| --- | --- | --- | --- |
| `sameSubjectTitle` | 0.000 | 0.000 | overlaps |
| Jaccard | 0.375 | 0.375 | ties exactly |
| Dice | 0.545 | 0.545 | ties exactly |
| **overlap coeff., stopwords dropped** | **0.667** | **0.545** | **separates** |
| title cosine (pinned model) | 0.628 | 0.715 | overlaps — worse than tokens |

The embedding approach is the one to reject: two unrelated rules from one domain read as similar
to the model. Swept at 0.60 with a 3-token floor over all 165,085 cross-repo pairs — 73 fire, 47
are auto-captured `Resolved failure in <path>` atoms excluded by prefix, 26 real subject pairs,
**8 promotion candidates**, containing all 7 known pairs plus one the original sweep missed.

**Ledger volume (does not pass).** Zero events; it shipped today. That is the half that keeps the
gate shut, and it closes only by using the workspace. Nothing here acts on the matcher — the
measurement exists so whoever builds the consolidator starts from a validated gate.

## 5 — a cross-repo evaluation suite, and a ranking change it killed

While measuring the above I built a "prefer local over a competing peer" ranking rule and had
strong-looking evidence for it: over one real three-repo workspace with everything shared, it took
3-of-12 changed top results down to 0-of-12. It broke two of `cross-repo-suite.json`'s three
cases the first time it met a fixture I had not written.

That is the actual gap this PR closes. Three cases over one two-repo fixture cannot tell you
whether a ranking change generalises, and my evidence was not a small sample but a **closed
loop** — the queries, the labels and the counterexamples all came from me, so the queries chosen
were ones where the local repo happened to own the answer.

`docs/evals/cross-repo-archetypes.json` adds **92 cases over 5 workspace archetypes, 15 repos,
121 items**: a split monorepo, diverged forks, unrelated client projects beside a house toolkit,
an asymmetric big-service/small-SDK/handbook trio, and four independent services. Ground truth was
authored **blind** — from "which item genuinely answers this developer?", with authors explicitly
told not to reason about which repo an item lives in. An independent audit of all 92 labels found
**zero broken cases**.

Two separations it holds deliberately:

- **Privacy is not a ranking metric.** A peer's repo-private row must never be returned; that is
  asserted at exactly zero from the fixtures' own visibility, not from `mustNotReturn` (which
  mixes the guarantee with "shouldn't crowd the page"). Merged, a ranking regression could pass as
  a privacy failure, or a leak hide inside a ranking tolerance.
- **Scored per archetype, never pooled** — a change that lifts the average while wrecking one
  shape is the failure this exists to catch.

Run against it, my rule loses **ten cells out of ten** (5 archetypes × 2 paths), MRR down 0.06 to
0.23, with `cross-repo-suite.json` agreeing independently at 1.0000 → 0.8333. Recall@3 barely
moves, which locates the damage exactly: correct answers get pushed off rank 1. The blind labels
keep putting the honest answer in a peer — a small SDK repo asking the platform, a service asking
shared infra — which is what a workspace is for.

**So the rule is not in this PR.** It is written up as rejected, with its numbers, in
`docs/evals/cross-repo-local-preference.md`, alongside two weaker variants rejected on the way.
What ships is the suite that killed it.
## Verification

`vitest run`: 628 suites / 2025 tests, 0 failures. `tsc --noEmit` clean. `eslint .` 0 errors
(26 warnings, all pre-existing). `check:lockfile`, `audit:prod`, `git diff --check` clean.

New tests: `federated-abstention` (5), `federated-kin` (4), `query-lineage-and-miss` (6),
`demand-ledger` (11), `demand-wiring` (4). The abstention tests were checked against the pre-fix
source by stashing it — 2 of 5 fail there, the 3 controls pass either way.

`docs:check` passes.


dat999zx:
Reviewed and merged main in. Three read-side fixes are solid — the floor plumbing in `queryFederated` mirrors `rankKnowledge` exactly, and the claim checks out: `NO CONFIDENT MATCH` really was unreachable from the moment a repo was linked. The kin marker is right to read sameness from the manifest rather than from a peer's own declaration, and the tests carry real controls (the null-floor case proving the labels come from the floor and not from weak queries) rather than just happy paths.

Pushed two commits: the merge of main (3.3.0 shipped while this was open, so the CHANGELOG entries move to `Unreleased`), and one round of review fixes. What changed in your code:

**The ledger's score column recorded a number that no longer means anything.** This is the one that mattered. `top_score` exists so a "weak query" threshold can be picked off its distribution, which needs a value meaning the same thing on every row — and `finalScore` stopped being one, twice over. `4152c34` (landed on main after you branched) min-max scales the semantic half across the candidate page, so the best row sits near 1.0 whether its cosine was 0.9 or 0.2; and with vector off the lexical half is normalised against each corpus's own best hit, landing near 1.0 again. `scoreOf` already refuses to publish that number to a caller for exactly this reason, so the ledger was recording what that gate rejects. It now stores the best raw cosine — the quantity the relevance floor is measured against, which makes a cut chosen here comparable to the shipped per-model floors.

A test caught a third case on the way: filtering only on "vector ran" isn't enough, because an unembedded row's semantic half is 0 by *absence*, not by verdict. A store whose background embedding hasn't caught up would have filled the column with fake misses. Uncalibrated rows are now excluded, and no calibrated row means no number at all.

**`openDemandDb` leaked a connection per concurrent write.** It cached the resolved client, so the whole bootstrap was a window where the cache is still empty — and since the writes are `void`-ed off a response path, a burst that all misses together is the normal traffic shape, not an edge case. Only the last connection stayed reachable; `closeDemandDb` couldn't reach the rest, and on Windows each holds its `-shm` sidecar for the life of the process. Now caches the in-flight promise. Your `takes concurrent writers on one file` test drove this exact scenario but asserted `total > 0`, which passes identically with one connection or twenty — it now asserts all 20 land, which is the number that distinguishes them.

**`summarizeDemand` created the ledger it was asked to report on**, because opening a libSQL `file:` URL creates the file. Same shape as the `doctor` fix in 3.3.0. An absent ledger is an answer, so it now reports empty and writes nothing.

Also dropped `demand_declines` (created on every open, read by nothing) and four unnecessary eslint-disable directives, which puts warnings back at main's 26.

**On the archetype suite — it caught a real regression on arrival**, and I want to be explicit that I did not quietly lower a number to make it green. `polyglot-services` semantic MRR goes 1.0 → 0.975 against 3.3.0. Traced by experiment rather than inference: neutralising `rescaleSemantic` restores 1.0 and leaves every other cell byte-identical, and the whole positional column never moves — which is what a semantic-only cause predicts. The case is `billing-wrong-timeout-applied`, a query that names a peer service, where amplifying a narrow cosine lead is the rescale working as designed.

I re-recorded it as an accepted cost with the cause, mechanism and experiment written into a `semanticRescaleCost` field, rather than reverting `4152c34`. That change won over 135 and 56 cases on the suites that can discriminate, and overturning it on a single fixture case is precisely the closed-loop reasoning your own `cross-repo-local-preference.md` exists to warn against. The invariant survives: the next drop in any cell is still a regression.

One thing I noticed and left alone: `knowl query` from the CLI is also a federated query and records no demand — only the MCP path and transcript search do. Defensible, and the plan states it, but the ledger's picture of "what one repo asks another" has a human-shaped hole in it. Worth a follow-up rather than a change here.

Verified on the merged tree: vitest 253 files / 2206 tests green, `tsc --noEmit` clean, eslint 0 errors, `docs:check`, `check:lockfile`, `audit:prod` and `git diff --check` all clean. Also smoke-tested `knowl workspace demand` end to end against a throwaway two-repo workspace, since the command has no test coverage.

Issue #35, opened 2026-08-07
linked-mode workspaces default to private visibility, which strands ~95% of knowledge to defend a boundary that has no counterparty


williamttruong:
## The observation

In `linked` mode, `visibility` has no counterparty.

- The knowledge database never leaves the machine. `.knowl/` is gitignored, and `doctor --fix`
  actively installs that entry (`installKnowlGitignoreEntry`), so a repo cannot carry its store to
  another person even by accident.
- Manifests hold local filesystem paths, and `knowlHome()` is one user's home directory.
- `shared` mode — the one where several people meet — is documented in `manifest.ts` as arriving
  in v2 and does not exist yet.

So in the shipped local product, the only reader of repo A's knowledge is the same person who
writes repo B's. `visibility: 'repo'` is enforced properly — the predicate is in the SQL, so a
peer's private row is never read into the querying process at all, which is the right way to build
it — but in `linked` mode it is a wall with nothing on the other side.

## What the default costs

`defaultVisibility` absent means `'repo'`, so every write is private unless someone remembers
otherwise, and `promote` is per-item with no inverse. That combination only drifts one way.

Measured on a real three-repo workspace:

| repo | private | shared |
| --- | --- | --- |
| A | 634 | 31 (4.7%) |
| B | 223 | 18 (7.5%) |
| C (has `defaultVisibility: workspace`) | 0 | 48 (100%) |

Nobody chose 95%. It accumulated. The visible symptom was an asymmetric surface: seven rules
existed word-for-word in both code repos, and in all seven the copy in one repo was shared while
its twin was private — an artifact of which repo someone happened to be standing in when they
thought to run the command.

## What sharing everything costs

Measured on a 92-case cross-repo suite spanning five workspace archetypes (added in #34), the same
cases run twice on the semantic path — once at the authors' chosen visibility, once with every
item forced workspace-visible:

| archetype | curated MRR | shared MRR | Δ | curated R@3 | shared R@3 |
| --- | --- | --- | --- | --- | --- |
| asymmetric-trio | 1.0000 | 1.0000 | — | 1.0000 | 1.0000 |
| monorepo-split | 1.0000 | 1.0000 | — | 1.0000 | 1.0000 |
| polyglot-services | 1.0000 | 1.0000 | — | 1.0000 | 1.0000 |
| fork-siblings | 0.9444 | 0.9167 | −0.028 | 0.9444 | 0.9444 |
[...]

PR #37, opened 2026-08-07
feat(workspace): a repo joining a linked workspace shares by default

williamttruong:
Closes the half of #35 you agreed to. Branched off `main` after #34 merged.

`workspace add` on a **`linked`** workspace stamps `defaultVisibility: workspace` unless told
otherwise, and prints one line above the existing gate notice saying a default decided it and how
to decline.

## Why

`'repo'` was the *compatibility* default — it preserved pre-workspace behaviour when the columns
landed and nothing read them, exactly as `4328408` says. Once workspaces shipped it became a
*policy* default nobody chose, and since promotion has no inverse it only drifts one way: measured
on a real three-repo workspace, to 95% private, with the same rule shared from one repo and
private in its sibling purely by where someone happened to be standing when they thought to run
`promote`.

Sharing costs little and loses nothing. Across the 92-case archetype suite, pooled MRR moves
0.9837 → 0.9674 while **recall@3 is unchanged to four decimal places** — answers get reordered,
never dropped, and three of five workspace shapes do not move at all
([`docs/evals/share-everything.md`](docs/evals/share-everything.md)).

That doc also gains the result from your review: the suspicion that per-commit capture noise
drives the delta, tested by injecting 120 such items and re-measuring. Identical to four decimals,
same forbidden count — noise never ranks for a real question, so it never enters the per-repo
candidate pool, and `DEFAULT_PER_REPO_CAP` means the pool is what matters rather than the corpus.
Excluding `fact`/`state` does recover about a third, but by sharing fewer *substantive* items, not
by withholding noise. So the narrower "keep the noise categories out" claim is not supported and
the doc says so.

## What it deliberately does not do

**Repos already in a manifest do not move.** An absent `defaultVisibility` still resolves to
`'repo'`. Changing what omission means would publish every linked repo's next write on account of
a release rather than a decision, with no demote to undo it — the bulk publish
`tests/cli/upgrade.test.ts` already forbids. The default applies only where a person ran a command
and read the notice. Pinned by a test that links one repo declining the default (which writes the
field absent, the same shape every pre-existing entry has) and then links another.

**The default does not satisfy `--promote-existing`.** That still requires an explicit
`--default-visibility workspace`.

This one is worth calling out because I got it wrong first. Resolving the default before the guard
made `workspace add ws --promote-existing` succeed silently — the guard compared the resolved
value, the new default satisfied it, and the largest irreversible action here would have run with
nobody having typed a visibility at all. The existing test caught it, which is the argument for
encoding a rule as behaviour rather than prose. It now checks the *requested* value, with its own
test so the two cannot be merged again for naming the same enum.

## The `linked` condition

Ships unconditional today — nothing constructs `'shared'` — and the CHANGELOG says so rather than
describing it as a limit. It is written now because `visibility` is one column on the item, not a
per-workspace grant: under `linked` a shared row means "me, in another directory"; under `shared`
the identical bit would mean "another person". Rows written under this default predate any
counterparty, so a future `shared` migration has to ask again rather than inherit them as consent.
A test pins the intent.

## Verification

`vitest` 663 files / 2217 tests green on the pre-rebase run and re-verified here, `tsc --noEmit`
clean, `eslint` 26 warnings (main's number), `docs:check`, `check:lockfile`, `audit:prod` and
`git diff --check` clean.

One expectation moved with the default: `workspace-cli`'s settings readout now shows `workspace`
after a plain `add`, annotated with why.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -101,7 +101,8 @@ Casual conversation, a single memory lookup, and trivial non-resumable work do n
 
 - When this repo is in a workspace, `knowl_query` results carry a `repo` field naming the repo that produced each item. A fact from another repo describes **that** repo unless it says otherwise; do not apply it here without checking.
 - Restrict a search to one repo with `knowl_query` `repos: ["<name>"]`. It matches the repo that owns an item.
-- Knowledge stays private to its repo until someone runs `knowl workspace promote`. Only the owning repo can promote, update, or retire its own items.
+- Whether a new write is shared is this repo's recorded default, not a fixed rule: joining a linked workspace sets it to workspace visibility, while `--default-visibility repo` keeps writes private. `knowl workspace set` with no flags prints the current value.
+- Knowledge already written privately stays private until someone runs `knowl workspace promote`. Only the owning repo can promote, update, or retire its own items.
 
 ### Safety and freshness
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,15 @@ is written now because `visibility` is one column on the item, not a per-workspa
 "another person". Rows written under this default predate any counterparty, so a future `shared`
 migration must ask again rather than inherit them as consent.
 
+Everything that told an agent or a reader the opposite moved with it. The "N existing items are
+still private" notice suggested re-running `add --promote-existing`, which the guard above now
+refuses — it names `--default-visibility workspace --promote-existing`, and a test parses the line
+the tool prints and runs it rather than asserting its wording. The guidance installed into every
+repo's `KNOWL.md` said knowledge stays private until someone promotes it; it now says visibility is
+the repo's recorded default and names the command that prints it, because that sentence is what an
+agent consults before deciding a write is safe. The README and `docs/reference.md` said the same
+thing on the front page and did not document the flag at all.
+
 ### A cross-repo evaluation suite that spans more than one workspace shape
 
 `cross-repo-suite.json` held **three** cases over a single two-repo fixture. That is not enough to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,40 @@ Notable changes to `@dat999zx/knowl`. Versions before 2.1.0 predate this file; s
 
 ## Unreleased
 
+### `workspace add` shares by default in a linked workspace
+
+**Breaking-ish, for new links only.** A repo joining a `linked` workspace now gets
+`defaultVisibility: workspace` unless it passes `--default-visibility repo`. The command says a
+default decided it and how to decline.
+
+`'repo'` was the *compatibility* default — it preserved pre-workspace behaviour when the columns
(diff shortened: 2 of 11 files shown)


dat999zx:
Accepted, with three fixes pushed to your branch as `9bd97ea` rather than sent back as a list — they are corrections to things the default made false, not disagreements with it.

It answers the comment on #35 on every point: the default applies where a person ran the command, existing entries never move, `--default-visibility repo` is the opt-out, the mode branch is written without being called a limit, and the shared-migration-must-ask-again rule is pinned. And you re-ran the measurement I asked for and reported that it does not support my hypothesis, which is the more useful outcome — the doc is right to refuse the category rule on a mechanism that measurement rules out.

## The one real defect: the two halves contradict each other

Your `--promote-existing` catch was correct, and it collides with the other half of the same PR.

Defaulting the visibility makes `existingItemsNotice` reachable **with no flags typed** — that notice used to fire only after someone had already passed `--default-visibility workspace`. Requiring an *explicit* flag for `--promote-existing` makes the short form of its advice fail. So on a plain `workspace add` in a repo holding knowledge:

```
1 existing item is still private.
Share them with:
  knowl workspace promote --category "fact,decision,..." --apply
Or re-run add with --promote-existing to do it in one step.
```

and doing exactly that:

```
Error linking repo: --promote-existing only applies with an explicit --default-visibility workspace, because it publishes everything this repo already knows.
```

Reproduced on a scratch workspace before anything was changed. The guard is right; the notice is what has to move, and it now names both flags. The test does not assert its wording — it parses the line the tool prints and runs it, so the guard and the advice cannot drift apart again for the same reason they did here. It fails on `e646cbb` with that error text.

## The one that matters more: the guidance said the opposite

`src/core/knowl-guidance.ts` still carried:

> Knowledge stays private to its repo until someone runs `knowl workspace promote`.

That block is installed verbatim into every repo's `KNOWL.md` and `AGENTS.md`. It is the sentence an agent reads before deciding a write is safe, so for a repo linked under the new default it was wrong in the publishing direction — the same shape as the `--promote-existing` hole, on the surface that reaches an agent every session rather than a person once. It now states that visibility is the repo's recorded default and names `knowl workspace set` as the way to read it. Both managed blocks regenerated from source rather than hand-edited, since they are rendered output.

The README made the same promise in three places, including the workspaces `<summary>` line, and `docs/reference.md` documented neither the default nor the flag on its `workspace add` rows.

## Small

`requested ?? (defaultedToWorkspace ? 'workspace' : undefined)` — the left side is already known `undefined` inside that branch. And a blank line between your default notice and the gate notice: run together they read as one paragraph and the irreversibility warning stops looking like a warning.

Verified on the pushed head: `tsc` clean, `eslint` 0 errors / 26 warnings (main's number), `docs:check`, `check:lockfile` clean, full suite **254 files, 2214 passed, 4 skipped**.

Merging once CI is green.

PR #45, opened 2026-08-09
feat(gc): record why an item was destroyed, not just that it was

williamttruong:
## The problem

GC computes a precise reason for every candidate — `Duplicate of <id>`, `State item stale for 47 days and never retrieved` — reports it in the run result, and then drops it. `deleteKnowledgeItem` writes the tombstone with the hardcoded literal `'purged'`, so **every purge in every store says the same word**. The deciding numbers exist for the length of one function call.

That leaves a collection policy unfalsifiable after the fact. You cannot ask which items were taken while they were still being retrieved, and you cannot retune a threshold against what it actually did, because nothing remembers what it did.

## What this adds

`knowledge_forget_log` — one append-only row per destroyed item, recording what was true at the instant of destruction: policy, reason, tier, status, retrieval count, last-retrieved time, age, size.

### Why not columns on the tombstone

`knowledge_tombstones` answers a different question and has to stay small: it is the record a delete **travels** on, written into every portable export (`portability.ts`) and merged by a monotonic upsert on import. Usage numbers there would push local retrieval telemetry into every export, and the upsert would let a peer's import overwrite this machine's audit trail with its own — or with nulls.

So the sync record and the audit record are separate tables, and only one leaves the machine. A test pins that: the tombstone appears in the export, the reason and the counts do not. The exported `'purged'` literal is deliberately left alone — widening it would change what leaves the machine.

### Two decisions worth checking

**No foreign key to `knowledge_items`.** The row it describes is already gone when this is written, and `ON DELETE CASCADE` would destroy the record at the exact moment it became the only copy.

**Snapshot policy `preserved`, not `restored`.** A restore cannot un-delete anything, so the record still holds — and `restored` would refill the table from the snapshot, erasing every deletion made since. That is precisely the window someone asking "was that threshold right?" is asking about. An audit trail a restore can silently truncate is not one.

**Pruning on request only**, unlike `pruneTombstones`' 90 days on every GC run. A tombstone older than any plausible export round has no job left; a forget-log row's job is to still be there months later when the question finally arrives. One row per destroyed item is a rounding error next to the item bodies GC exists to reclaim.

## A defect the tests did not catch

Verified against a copy of a real 748-item store. GC found nothing to collect there, so I forced the path by deleting the three **most-retrieved** items — and got three rows all reading `retrievals=0`, because the no-context delete path defaulted the count instead of reading it.

A log that states a false number is worse than one that states none: it reads as evidence the item was dead. `deleteKnowledgeItem` now reads the real count when the caller supplies none; GC supplies its own whole-store summary and is preferred. Same three items after the fix: **39, 44 and 56**. Both behaviours are pinned by tests.

## One change outside the feature

`impact-schema.test.ts`'s migration-level assertion moves from `=== 4` to `>= 4`. It pinned a global constant from inside one feature's test, so any later additive change anywhere in the store fails an impact test that has nothing to do with it — which is how it failed here. `schema-pin.test.ts` already pins the exact level against the real schema hash, so the value keeps an owner that cannot drift. Say the word if you would rather I bumped the literal instead.

Migration level 5, additive: one new table, no altered column, `KNOWL_SCHEMA_VERSION` unmoved. A backfill is not merely skipped but impossible — the deciding numbers for every past deletion were discarded at the moment of deletion.

## Verification

- Full suite **267 files, 2,367 tests, 0 failures**; `npm run lint` 0 errors, 27 warnings (unchanged baseline, none in touched files); `npm run typecheck` clean.
- 8 new tests: reason and evidence recorded, outlives the tombstone prune, never appears in an export, survives the item with no cascade, logs a context-free delete rather than nothing, reads the real retrieval count, prefers caller-supplied numbers, newest-first ordering and limit.
- Schema-pin and snapshot-table-ownership gates both satisfied rather than bypassed.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -765,6 +765,26 @@ Flags adjust these thresholds, and `--ignore-access` disregards access heat for 
 archival. Review the preview before applying it; GC does not infer semantic equivalence across
 different content.
 
+### The forget log
+
+Every destroyed item leaves one append-only row recording what was true at the instant of
+destruction: the policy that fired, its reason in words, and the retrieval evidence that policy
+decided against. That makes a collection threshold checkable after the fact — you can ask which
+items were taken while they were still being retrieved.
+
+```bash
+knowl forget-log
+knowl forget-log --limit 100 --json
+knowl forget-log --repo web            # only items owned by that workspace repo
+knowl forget-log --prune-days 365
+```
+
+This is a separate table from `knowledge_tombstones` and it never leaves the machine. A tombstone
+rides in every portable export and merges by upsert on import, so retrieval numbers there would
+both publish local telemetry and let a peer's import overwrite this machine's audit trail.
+Tombstones are pruned after 90 days on every GC run; forget-log rows are kept until
+`--prune-days` asks for them to go, because the question they answer arrives months late.
+
 ### Snapshots, audit, and doctor
 
 ```bash
@@ -1233,6 +1253,7 @@ knowl eval retrieval --dataset docs/evals/retrieval-suite.json --json
 | `knowl reindex --transcripts [--budget <minutes>]` | Build or update the optional session transcript index; resumable, so a budget is a stopping point rather than a rollback |
 | `knowl resume [key]` | Resume a parked workstream from its key, or list what is parked here |
 | `knowl gc [--apply] [--stale-days N] [--compress-days N] [--min-bytes N] [--ignore-access] [--tombstone-days N]` | Preview or apply duplicate, archive, compression, and tombstone maintenance |
+| `knowl forget-log [--limit N] [--repo <name>] [--json] [--prune-days N]` | Show why knowledge items were destroyed — policy, reason, and the retrieval evidence it overruled — or prune those records |
 | `knowl pr check --since <commit> [--dry-run]` | Find drift candidates and, unless dry-run, mark them for review |
 | `knowl view [--port <port>]` | Start the local GET-only viewer |
 | `knowl serve` | Start the stdio MCP server |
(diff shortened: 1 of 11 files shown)


williamttruong:
Pushed a follow-up (`0f9776a`) after reading Lethe's actual `forget_log.py` rather than the README paragraph the first pass was built from. Three things its implementation has that this PR did not, none of them visible in the README:

**Reason codes as a closed set.** Lethe keeps `REASON_COLD_EXPIRED` / `REASON_MERGED` as module constants and states why: they are *"part of the public ForgetLog contract — anyone reading the log needs to know what the strings mean."* A log whose only reason is prose cannot be aggregated; "how many were merged versus collected cold" should be a `GROUP BY`, not a regex over English. `reason_code` is now that closed set, beside the human sentence rather than replacing it.

**The survivor as a column.** Duplicate collection recorded the winner only inside the sentence `Duplicate of <id>`, so following it meant parsing it back out. `merged_into_id` makes "what was merged into X" answerable — and GC already carried `duplicateOfId` on the candidate, it was just discarded.

**A bounded content snapshot.** Lethe stores content because *"the item itself may have been physically removed by the time anyone reads the log"* — exactly the case here, since a purge is a hard delete. Judging whether a policy was right usually means seeing what it destroyed. Bounded at `MAX_PREVIEW_CHARS` so the audit trail cannot quietly undo the reclamation the collection was for. The commit log still holds the whole `before` item; this is the copy in the row someone actually reads.

Deliberately **not** taken: Lethe's `ForgetLog` ABC with pluggable in-memory and SQLite backends. That exists because its benchmark harness needs both. Knowl has one store and the indirection would buy nothing.

Level 5's pin is amended rather than a level 6 added, since level 5 has not shipped and its schema is still being defined.

Full suite **267 files, 2,370 tests, 0 failures**; lint 0 errors (27 warnings, unchanged baseline); typecheck clean. Three new tests cover the three additions.

PR #48, opened 2026-08-09
feat(gc): a groupable rule code, the survivor as a column, and what was destroyed

williamttruong:
Follow-up to #45. I had built that PR from a paragraph of Lethe's README; this is what its actual `forget_log.py` (MIT) carries that the table did not. Reading the implementation first would have put these in the original.

It also builds on `2466d14` rather than around it — the reader you added is what makes two of these three worth having at all.

## What this adds

**`reason_code`, a closed set beside the sentence.** Lethe keeps its reasons as module constants and says why: they are *"part of the public ForgetLog contract — anyone reading the log needs to know what the strings mean."* `reason` alone is prose, and prose cannot be aggregated: "how many were merged versus collected cold" should be a `GROUP BY`, not a regex over English. A new collection rule now adds a code rather than another ungroupable sentence.

In the text listing the sentence still leads and the code follows it. Your existing format assertion was right and I initially broke it — a tally happens over `--json` or SQL, where the field already is; a human reading the list wants the sentence that names the item.

**`merged_into_id`.** Duplicate collection recorded the survivor only inside `Duplicate of <id>`, so following it meant parsing it back out. GC already carried `duplicateOfId` on the candidate and dropped it. `knowl forget-log` now prints it.

**`content_preview`.** Lethe stores content because *"the item itself may have been physically removed by the time anyone reads the log"* — exactly the case here, since a purge is a hard delete. Bounded at `MAX_PREVIEW_CHARS` so the audit trail cannot undo the reclamation the collection was for; `bytes` still records the real size, so bounding the preview loses nothing about cost. The commit log still holds the whole `before` item — this is the copy in the row someone actually reads.

Deliberately **not** taken: Lethe's `ForgetLog` ABC with pluggable in-memory and SQLite backends. That exists because its benchmark harness needs both; Knowl has one store.

## Level 7, and the ALTER pass that makes it real

A level of its own rather than an amendment of 6, for the reason 6 could not amend 5: **6 has shipped**, so a store already stamped at it skips `SCHEMA_STATEMENTS` entirely.

That has a second consequence I nearly missed. `CREATE TABLE IF NOT EXISTS` is a no-op on a store that already has the table, so the columns would have existed only on databases created after this change — and every older store would fail its next delete, inside the single transaction `applyKnowledgeGc` runs, taking the whole collection run down. `ensureForgetLogColumns` adds them by `ALTER`, in the same shape as `ensureQualityColumns`.

Nothing is backfilled: a row written at level 6 has no survivor and no snapshot, and `reason_code` falls to `'unspecified'`, which is the honest description of a deletion recorded before the codes existed rather than a guess at which rule fired.

## Verification

- Full suite **268 files, 2,392 tests, 0 failures**; lint 0 errors, 27 warnings (unchanged baseline); typecheck and build clean.
- Five new tests: the code and survivor as columns, the snapshot, its bound, the default code on a context-free delete, and the **level-6 → 7 upgrade path** — which rewinds `application_id` first, because `bootstrapSchema` skips every pass when the stamp already reads current, and a migration test that cannot fail proves nothing.
- The upgrade test was mutation-checked: removing the `ensureForgetLogColumns` call fails it.

### DIFF
--- a/src/cli/program.ts
+++ b/src/cli/program.ts
@@ -57,6 +57,7 @@ import { parseAgentNames } from './agents/registry.js';
 import { reindexKnowledgeEmbeddings } from '../store/vector-index.js';
 import { applyKnowledgeGc, previewKnowledgeGc, isHot } from '../store/gc.js';
 import { listForgetLog, pruneForgetLog } from '../store/forget-log.js';
+import { truncateText } from '../core/token-budget.js';
 import { getAccessSummary } from '../store/access-feedback.js';
 import { checkpointWorkLoop, finishWorkLoop, startWorkLoop, WorkLoopMemoryHit } from '../store/work-loop.js';
 import { checkKnowledgeDrift, DriftCheckResult, getCurrentGitCommit, listChangedFilesSince } from '../store/drift.js';
@@ -2155,7 +2156,16 @@ program
       for (const entry of entries) {
         const owner = entry.originRepo ? ` [${entry.originRepo}]` : '';
         console.log(`- ${entry.deletedAt} ${entry.policy} ${entry.itemId}${owner} ${entry.title}`);
-        console.log(`  Reason: ${entry.reason}`);
+        // Sentence first, code after. The code is what a tally groups by, but that happens over
+        // `--json` or SQL, where the field is already there; a human reading this list wants the
+        // sentence that says which item this was.
+        const survivor = entry.mergedIntoId ? ` -> ${entry.mergedIntoId}` : '';
+        console.log(`  Reason: ${entry.reason} [${entry.reasonCode}]${survivor}`);
+        // The body, because the item is gone: judging whether a rule was right usually means
+        // seeing what it took, and this row is the only copy anyone reads.
+        if (entry.contentPreview) {
+          console.log(`  Was: ${truncateText(entry.contentPreview, 160)}`);
+        }
         // The retrieval numbers are the point: a purge of something still being read is the
         // finding a threshold review is looking for, and it is invisible in a plain count.
         const lastSeen = entry.lastRetrievedAt ? `, last ${entry.lastRetrievedAt}` : '';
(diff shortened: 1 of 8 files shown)


dat999zx:
Rebased onto `main` (`8969463`) and moved to **migration level 8**. Force-pushed to this branch; your authorship is preserved on the commit.

## Why it had to move

`cloud_published` shipped as level 7 in #50 while this branch was open, so both claimed 7 — with different fingerprints (`b847bdb2…` here vs `154d43b0…` there). That is worse than an ordinary merge conflict, because **both sides compile and both stamp the same `application_id`**. Whichever migrates a store first tells the other's gate there is nothing to do, and the second set of columns is then silently never created on that store — no error, no missing-table exception until the next delete fails inside `applyKnowledgeGc`'s transaction.

Your own header comment argued exactly this for 6-not-5. It applies once more.

## What changed in the rebase

- `KNOWL_MIGRATION_LEVEL` 7 → **8**; your level-7 block became the level-8 block and now sits after main's `cloud_published` entry rather than replacing it.
- `SCHEMA_PINS[8]` recomputed against the real level-8 schema: **`e96218e9fbe0594d73078033d718af4f`**. The old `154d43b0…` was the fingerprint of a schema without `cloud_published`, so it could not have been carried over.
- Added a line recording *why* it is 8 and not 7, so the next branch in this position does not have to rediscover it.

Nothing else was touched. `ensureForgetLogColumns`, the `reason_code` closed set, `merged_into_id` and the bounded `content_preview` are all as you wrote them.

## Verification after the rebase

| command | result |
|---|---|
| `npm run typecheck` | clean |
| `npm run lint` | 0 errors, 30 warnings (baseline) |
| `npm run build` | clean |
| `npm test` | 297 files, 2658 passed, 5 skipped, **0 failures** |
| `schema-pin` + `forget-log` + `forget-log-command` | 28 passed |

One note on the review: the `ALTER` pass is the right call and `NOT NULL DEFAULT 'unspecified'` is legal for SQLite's `ADD COLUMN`, so a level-6 store widens cleanly. The one thing I'd still question is the code set itself — `'delete:unspecified'` for a manual delete and `'unspecified'` for a pre-level-8 row are one character apart in meaning and easy to confuse in a `GROUP BY`. Not blocking; worth a rename if you agree.

PR #77, opened 2026-08-11
fix(tests): carry the git identity on the invocation, so a fixture cannot rewrite this repo's config

dat999zx:
> **Stacked on #76.** Base is `fix/archive-reader-silent-losses`, so the diff here is only this
> change. `scripts/e2e-cloud.ts` is touched by both, which is why this is stacked rather than
> branched from `main` — GitHub will retarget the base to `main` automatically when #76 merges.

## The bug this closes bit us while opening #76

`git config user.email X` run against a fixture directory can write into **this** repository.
`git config` with no `--file` searches upward for the repository the directory belongs to, so
when the fixture has no `.git` at that instant — the window between `git init` and the next
line, or a run interrupted before it — the write lands on the nearest enclosing repository.
Fixtures here are routinely built at `path.resolve('./.knowl-...')`, which is inside knowl.

It happened, and nothing caught it. This repository's `.git/config` held:

```
[user]
	name = Test
	email = test@example.com
```

— matching [drift-auto.test.ts:45-46](tests/store/drift-auto.test.ts#L45-L46) exactly. The next
commit went out authored by it, GitHub resolved `test@example.com` to an **unrelated person's
account**, and the `cla` check failed correctly against a stranger who had never signed it. The
commit had to be re-authored and force-pushed.

A local `user.email` silently overrides the developer's global one for every later commit, so
nothing about the working copy looks wrong afterwards. That is what makes it worth a guard
rather than a note.

## The fix, in two commits, and the second is the interesting one

**`a4a3d9c` — carry the identity on the invocation.** `git -c user.name=... -c user.email=...`
sets the value for one process and writes no file, so there is no file to land in the wrong
place. 26 call sites across five `tests/cloud/` suites, six `tests/store/` suites,
`tests/cli/cli.test.ts`, and both `scripts/e2e-*.ts`. The pattern was already here —
[index-pass.test.ts:205](tests/transcripts/index-pass.test.ts#L205) has always done it this way.

One identity for all of them rather than the nine each suite had invented, because **no
assertion anywhere reads one** — checked before collapsing them.

**`700b984` — put it in the environment, because the first commit was not enough, and only CI
could tell.** Converting each suite's helper covers the call sites you edited. It missed a
second, raw `execSync('git commit ...')` sixty lines below the helper in `cli.test.ts`. Every
local run passed, because a developer's global `~/.gitconfig` supplies an identity and the raw
call silently inherited it. CI runners have none:

```
fatal: empty ident name (for <runner@runnervmvrwv9...>) not allowed
```

Three legs — ubuntu 22, ubuntu 24, windows 22 — one test, one cause.

The lesson is about the shape of the fix rather than the one missed line: a convention every
call site must remember is one some call site will not, including sites not written yet. So the
identity now comes from `GIT_AUTHOR_*` and `GIT_COMMITTER_*` set in the vitest global setup,
beside the TMPDIR redirect and relying on the same property — workers inherit the parent
environment — so every existing and future git invocation is covered with no test-file change.
All four variables, because git requires AUTHOR and COMMITTER separately and refuses on either
alone.

The `-c` flags stay: they are what keeps `git config` out of the tree, and they are the only
mechanism available to `scripts/e2e-*.ts`, which do not run under vitest.

## The guard

`tests/architecture/git-identity.test.ts`, beside the module-boundary contract and for the same
stated reason: nothing was looking. Two halves:

1. **No `git config user.*` in `tests/` or `scripts/`.** Names the offending file and line
   rather than counting them (the fix is per line). Comment lines are exempt — every place this
   rule is explained has to name the thing it forbids.
2. **The identity reaches a worker.** Asserted in a worker rather than trusted from `setup`,
   because inheritance across the worker boundary is the property being relied on.

Verified to discriminate, not just to pass. Reintroducing a call:

```
tests/store/write-gate.test.ts:161: git('config', 'user.email', 'sneaky@example.test');
```

## Verification

`typecheck`, `lint`, `git diff --check` clean.

**Run under CI's condition rather than this machine's**, which is the check that was missing the
first time. With `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` pointed at an empty file, so no
global identity exists anywhere:

```
: > /tmp/empty-gitconfig
GIT_CONFIG_GLOBAL=/tmp/empty-gitconfig GIT_CONFIG_SYSTEM=/tmp/empty-gitconfig npx vitest run
→ 305 files, 2,741 passed, 5 skipped, 0 failed
```

The same command reproduces the CI failure on `a4a3d9c`, so it discriminates.

## Follow-up this surfaced, deliberately not in scope

Local full runs on this branch also hit `EPERM: operation not permitted, rename '<file>.tmp' ->
'<file>'` in files this PR never touches. That is the known Windows atomic-write flake, and it
has now crossed the threshold the earlier investigation set for it (*"more than roughly one run
in ten"*): two of three local full runs. **I initially mistook it for the whole story and missed
the identity regression underneath it** — worth saying, because the same reasoning error is
available to the next person.

The remedy that investigation named is a bounded retry around the rename in the atomic-write
helper — one place, since every call site goes through it — covering `src/core/config.ts:209`
and `src/cloud/credentials.ts:65`. It is a real product bug on Windows, not only test noise:
`saveConfig` throwing EPERM means an ordinary `knowl init` can fail. Left out because it changes
shipped code and is unrelated to git identity. Worth its own PR.

### DIFF
--- a/scripts/e2e-cloud.ts
+++ b/scripts/e2e-cloud.ts
@@ -58,18 +58,19 @@ async function main(): Promise<void> {
   const ORIGIN = path.resolve('./.knowl-e2e-origin');
   await fs.rm(ORIGIN, { recursive: true, force: true }).catch(() => {});
   await fs.mkdir(ORIGIN, { recursive: true });
-  const inOrigin = (args: string[]) => spawnSync('git', args, { cwd: ORIGIN, encoding: 'utf8' });
+  // Identity carried on each invocation rather than written with `git config`. These fixtures sit
+  // inside this repository, and `git config` searches upward when the fixture has no `.git` yet --
+  // so a run interrupted at the wrong moment leaves a bogus `user.email` in knowl's own config,
+  // and the next commit made here goes out authored by it. See `tests/git-identity.ts`.
+  const IDENTITY = ['-c', 'user.name=Knowl-E2E', '-c', 'user.email=e2e@example.test'];
+  const inOrigin = (args: string[]) => spawnSync('git', [...IDENTITY, ...args], { cwd: ORIGIN, encoding: 'utf8' });
   inOrigin(['init', '-q', '-b', 'main']);
-  inOrigin(['config', 'user.email', 'e2e@example.com']);
-  inOrigin(['config', 'user.name', 'E2E']);
   await fs.writeFile(path.join(ORIGIN, 'a.txt'), 'one', 'utf8');
   inOrigin(['add', '.']);
   inOrigin(['commit', '-qm', 'one']);
   spawnSync('git', ['clone', '-q', ORIGIN, ROOT], { encoding: 'utf8' });
 
-  const git = (args: string[]) => spawnSync('git', args, { cwd: ROOT, encoding: 'utf8' });
-  git(['config', 'user.email', 'e2e@example.com']);
-  git(['config', 'user.name', 'E2E']);
+  const git = (args: string[]) => spawnSync('git', [...IDENTITY, ...args], { cwd: ROOT, encoding: 'utf8' });
   // The refs are what the gate reads; the URL is only what identity is derived from. Rewriting it
   // after cloning gives a realistic `github.com/acme/e2e` without losing `origin/main`.
   git(['remote', 'set-url', 'origin', 'git@github.com:acme/e2e.git']);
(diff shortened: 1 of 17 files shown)


dat999zx:
Superseded by #78. This PR was auto-closed when its base branch `fix/archive-reader-silent-losses` was deleted on merging #76; GitHub will not reopen or retarget a closed PR, so the same two commits were rebased onto `main` and opened fresh.

The CI history here is worth keeping: the first run shows the `fatal: empty ident name` failure on ubuntu 22, ubuntu 24 and windows 22 that a local run could not reproduce, which is what motivated the second commit.

PR #123, opened 2026-08-16
fix(discovery): a git worktree reaches the repository's memory instead of reporting no project

williamttruong:
## The bug

`.knowl/` is gitignored, and `git worktree add` materialises tracked files only. So a linked worktree is a checkout of an initialized repository that carries **no marker anywhere inside it**. Placed outside the main checkout — which is where orchestrators put them — the ancestor walk in `findProjectRoot` has nothing to reach.

Measured against released **5.4.0**, in a real worktree of a Knowl project:

```
$ git worktree add <tmp>/wt -b agent-1
$ cd <tmp>/wt && knowl doctor
[FAIL] No Knowl project found at "...\wt". Run "knowl init" to initialize.
Result: NOT READY

$ knowl query "..."
Error querying knowledge: No Knowl project found
```

Not a degraded result — a hard error on every command.

**This is the whole of the parallel-agent case.** `isolation: "worktree"` on Claude Code's Agent tool, this repository's own `docs/audit-lanes.md` (which runs seven audit lanes in hand-made worktrees), and every orchestrator that fans agents across isolated checkouts all produce exactly this shape. The agents that most need shared memory were the ones guaranteed not to have it — silently, because `No Knowl project found` reads as a setup mistake rather than a product gap.

Worth noting the failure is placement-dependent, which is why it stayed hidden: a worktree placed *inside* the main checkout already worked, because the ordinary walk climbs out into it.

## The fix

`mainWorktreeRoot` resolves the shared git directory via `git rev-parse --git-common-dir`. Measured output, which is why a single `path.resolve` covers all of it:

| run from | prints |
|---|---|
| main checkout root | `.git` |
| subdirectory of main | `../../.git` |
| linked worktree root | `C:/repo/.git` (absolute) |
| subdirectory of a worktree | `C:/repo/.git` (absolute) |
| outside any repository | exits non-zero |

Relative in the first two and absolute in the next two, so resolving against the start path normalises every case to the same shared `.git`, whose parent is the main checkout.

Only a directory literally named `.git` is accepted. A submodule reports `.git/modules/<name>`; there the parent is not a checkout, and guessing would hand back something that merely looks like one. Returns null when git is absent or the path is not in a repository — this widens a lookup that already failed, so it must not turn a missing git binary into a new error.

### What keeps it safe

- **It is a fallback, never a first step.** A Knowl project nested inside a larger git repository keeps resolving to itself, because the ordinary walk reaches its marker before this runs (K-09).
- **It cannot invent a project.** The second walk applies `isProjectRoot` exactly as the first did, so a repository whose main checkout was never initialized still throws rather than borrowing a store.
- **`findProjectRootSync` gets the same fallback.** It answers for `assertDatabasePresentForCommand`, which reads null as "not a project, nothing to check" — a resolver that gave up where the async one now succeeds would skip the missing-database guard in precisely the checkouts that had just started resolving.

## Verification

Six new tests in `tests/cli/project-marker.test.ts`, written failing first against a real `git worktree` fixture. They pin the premise (the worktree genuinely has no marker), both resolution paths, the sync resolver, and the two ways this could go wrong — a worktree whose main checkout has no marker, and a plain uninitialized checkout — both of which must still throw.

- `npm test` — **337 files, 3105 passed, 5 skipped, 0 failed**
- `npm run typecheck` — clean
- `npx eslint` on changed files — clean

End-to-end against a real worktree of a Knowl project, same command before and after:

```
BEFORE (5.4.0):  [FAIL] No Knowl project found ... Result: NOT READY
AFTER:           [OK] Repository initialized at C:\Code\knowl-cloud\.knowl
                 [OK] Config loaded
                 [OK] Cloud: github.com/dat999zx/knowl-cloud → ... synced
```

and `knowl query` from inside the worktree returns real atoms from the main checkout's store.

## Note on scope

This makes N worktrees of one repository share **one** store, which is the behaviour parallel agents want. Whether N *simultaneous* writers produce semantically coherent memory is a separate question and is not addressed here — WAL plus `busy_timeout` and the connection-pool retry mean concurrent writers get bounded waits rather than errors, but that is a durability property, not a coherence one.

### DIFF
--- a/src/cli/database-presence.ts
+++ b/src/cli/database-presence.ts
@@ -1,6 +1,7 @@
 import { existsSync, readFileSync, statSync } from 'node:fs';
 import path from 'node:path';
 import { resolveStorage } from '../store/storage-roles.js';
+import { mainWorktreeRoot } from '../core/config.js';
 import { repoRegistryPath } from './repo-registry.js';
 
 /**
@@ -54,8 +55,7 @@ function isProjectRootSync(candidate: string): boolean {
   }
 }
 
-/** The enclosing initialized repository, or null when there is none. */
-export function findProjectRootSync(startPath: string): string | null {
+function walkForProjectRootSync(startPath: string): string | null {
   let current = path.resolve(startPath);
   for (;;) {
     if (isProjectRootSync(current)) return current;
@@ -63,6 +63,24 @@ export function findProjectRootSync(startPath: string): string | null {
     if (parent === current) return null;
     current = parent;
   }
+}
+
+/**
+ * The enclosing initialized repository, or null when there is none.
+ *
+ * Falls back to the main checkout for a linked worktree, exactly as `findProjectRoot` does and
+ * for the same reason -- `.knowl/` is gitignored, so a worktree carries no marker. Kept in step
+ * deliberately: this guard exists to turn a vanished database into an explanation rather than a
+ * raw error, and a resolver that gave up where the async one succeeds would skip the guard in
+ * precisely the checkouts that had just started resolving.
+ */
+export function findProjectRootSync(startPath: string): string | null {
+  const direct = walkForProjectRootSync(startPath);
+  if (direct) return direct;
+
+  const main = mainWorktreeRoot(startPath);
+  if (main && path.resolve(main) !== path.resolve(startPath)) return walkForProjectRootSync(main);
+  return null;
 }
 
 /** Has this machine ever initialized or upgraded this repository? */
--- a/src/cli/program.ts
+++ b/src/cli/program.ts
@@ -329,14 +329,28 @@ program
       // is that same question asked one directory further up.
       const enclosing = await findProjectRoot(cwd).catch(() => null);
       if (enclosing && path.resolve(enclosing) !== path.resolve(cwd)) {
+        // Two different ways to get here, and the same words are false in the second. `cwd`
+        // under `enclosing` is the subpackage case above. `cwd` NOT under it means the resolver
+        // reached the project sideways, through the shared git directory of a linked worktree --
+        // so "inside", "run commands from anywhere under" and "move it outside" would each
+        // describe something that is not true of where the user is standing.
+        const relative = path.relative(enclosing, cwd);
+        const nested = relative !== '' && !relative.startsWith('..') && !path.isAbsolute(relative);
         throw new Error(
-          `${cwd} is inside the Knowl repository at ${enclosing}.\n` +
-          'Initializing here would create a second store, and every command run below this ' +
-          'directory would then read and write that one instead of the repository\'s -- ' +
-          'silently, because both are valid.\n' +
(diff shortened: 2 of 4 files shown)


dat999zx:
Reviewed and pushed three fixes to the branch (`7792be3`). The diagnosis and the shape of the fix are right — a worktree genuinely could not reach its own memory, and `--git-common-dir` is the correct question. Two things made the answer unsafe, and one was collateral.

**1. `GIT_DIR` outranks `cwd`.** `git rev-parse` honours `GIT_DIR`/`GIT_COMMON_DIR` over the working directory, and git exports them into every hook it runs, so any subprocess of `git commit` inherits them. Measured through the built CLI:

```
$ knowl status          # in a dir that is not a repo and not a worktree
❌ No Knowl project found at "...\unrelated\deep"

$ GIT_DIR=<repo>/.git knowl status      # same dir
Repository:     ...\main
```

It silently read and wrote an unrelated repository. The docstring said this returns null when "the path is not in a repository"; under an inherited `GIT_DIR` it did not.

**2. The miss branch is `agent-hook`s hot path.** `agent-hook.ts` says it best: it runs once per agent tool call as a fresh process, "the only Knowl command with that property", and names `ProjectNotFoundError` as the *ordinary* case because the hook fires in every directory the agent visits. A `git` spawn measures ~22ms — about 13% of the 175ms the bundling work bought, unamortisable because the process is new each time. Ordinary CLI commands in a worktree paid it twice, once in the `preAction` guard and once in the command body.

**One gate closes both.** `git worktree add` writes `.git` as a **file** (`gitdir: …`); a main checkout writes a **directory**. Not spawning git unless that marker is in an ancestor keeps the cost on the path that needs it, and a directory carrying an inherited `GIT_DIR` has no marker to find. I also stripped `GIT_DIR`/`GIT_COMMON_DIR`/`GIT_WORK_TREE` from the child env, so the gate is not the only thing between us and the wrong repository.

**3. Collateral: `knowl init` in a worktree.** The lookup now succeeds, so the enclosing-repository guard fires — with a message written for a nested subpackage and false on every line here:

```
❌ ...\trees\agent-1 is inside the Knowl repository at ...\main.
  - to use the existing memory: run knowl commands from anywhere under ...\main
  - if this really is a separate project, move it outside that repository first
```

It is not inside; you no longer need to be under `main` to reach the memory (that is what this PR fixed); and it is already outside. Refusing is still right, so the refusal now describes the case it is in.

Three tests added: inherited `GIT_DIR` resolves to nothing, a path with no worktree marker answers null without consulting git, and init in a worktree prints the worktree wording and still writes no second store.

Local: 15/15 in `project-marker`, `tests/cli tests/core` 804 passed / 0 failed, typecheck and eslint clean. Merging once CI is green.

Nice catch on the placement-dependence, by the way — that a worktree *inside* the checkout worked is exactly why this stayed hidden.

PR #124, opened 2026-08-16
fix(guidance): listed but not callable is not unavailable

williamttruong:
## What happened

A Claude Code session (2026-08-16) listed **every** knowl tool by name under:

> The following deferred tools are now available via ToolSearch. Their schemas are NOT loaded — calling them directly will fail with `InputValidationError`.

Reaching `knowl_query` first therefore cost a schema-load round-trip that our guidance never describes. The host is doing progressive disclosure *to* us, and we get no say in the subset or the ordering.

The sharp part is rule 6, which read:

> If Knowl MCP tools are unavailable, stop and tell the user instead of silently bypassing Knowl.

An agent staring at a list of uncallable tools can reasonably read that as its own case — and the instructed response to that case is **to stop using Knowl**.

## What this changes

**1. Rule 6 distinguishes deferred from absent** (`renderFullKnowlGuidance`). `KNOWL_SUBAGENT_BOOTSTRAP_CARD` already carries this line, because a subagent is dispatched into exactly that shape and a probe found it. Same failure in the main session, same sentence. The genuine-absence half is kept intact.

**2. It prints the name that resolves.** A namespacing host builds the callable name from the registration key, and `knowl init` writes `mcpServers.knowl` — so the namespaced form is `mcp__knowl__knowl_query`, which is what an agent has to select by *exact name* when schemas are withheld. The key is now one exported constant (`KNOWL_MCP_SERVER_KEY`) shared by the guidance and both config writers, with a test asserting the printed prefix against what the writers actually emit.

That join is the part worth having. A surface that *prints* an identifier drifting from the identifier that *resolves* has already bitten this repo twice with CLI command names (`knowl workspace connect`, `knowl cloud login`), both times because nothing connected the two halves.

**3. Doctor stops claiming "MCP remains available."** The degraded and unsupported lifecycle-hook messages promised the tool surface is a sufficient fallback. It is only sufficient on a host that loads tool *descriptions*; where schemas are deferred the routing prose is not in context at all, and the host instruction file plus the server card are the only carriers left. The messages now name what actually survives — the recording, not the routing.

## Two things I deliberately did not do

**Nothing lands in the compact cards.** The transcript-enabled card sits at **1,990 / 2,000**, and this sentence does not fit. The full guidance is a file written once per repo, not a cost paid per session, so it is the right home. Whether the card should buy the room by compressing an existing line is a real question, but it spends a contested budget and touches prose with an eval harness behind it — that is your call, not a drive-by.

**Doctor severity is unchanged.** Flipping hooks-stale from WARN to FAIL was my first instinct and I think it is wrong: deferral is a client-side presentation choice made *after* `initialize` and `tools/list`, so the server cannot detect it, and an unconditional FAIL would report NOT READY on every host where nothing is actually wrong. If you want the severity rule revisited, that is a separate decision with real blast radius — `ready` gates on it.

## Verification

- `tsc --noEmit` — clean
- `eslint .` — clean
- `npm run docs:check` — current (`KNOWL.md` / `AGENTS.md` regenerated)
- `npx vitest run` — **337 test files, 3,100 tests, 0 failures**

Diff is 6 files, +94/−11.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,7 +74,7 @@ Strong success criteria let you loop independently. Weak criteria ("make it work
 3. Use a relevant active hit immediately. Inspect files only after a miss, conflict, stale/low-confidence memory, or explicit verification request.
 4. Query again before switching to a distinct subtask or project area, and before choosing how to build something new — existing tooling and pipelines are project knowledge, and in a linked workspace they often live in a sibling repo, so leave method queries unscoped.
 5. Store or update durable knowledge during work and before the final answer — verified findings, stated intent (goals, plans, direction the user voiced) stored as goals with user_stated provenance even while unsettled, and resolved diagnoses stored as skills when the cause will recur (an environment quirk, a config trap — not a typo). The test: could a fresh session recover this from memory alone? Never store raw transcripts, secrets, or transient debugging noise.
-6. If Knowl MCP tools are unavailable, stop and tell the user instead of silently bypassing Knowl.
+6. Listed but not callable is not unavailable. A host may namespace the tools and withhold their schemas until asked, so load the schema for the name you need — namespaced it is `mcp__knowl__knowl_query` — and call it. Only when the tools are genuinely absent, stop and tell the user instead of silently bypassing Knowl.
 
 ### Lifecycle modes
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,7 +74,7 @@ Strong success criteria let you loop independently. Weak criteria ("make it work
 3. Use a relevant active hit immediately. Inspect files only after a miss, conflict, stale/low-confidence memory, or explicit verification request.
 4. Query again before switching to a distinct subtask or project area, and before choosing how to build something new — existing tooling and pipelines are project knowledge, and in a linked workspace they often live in a sibling repo, so leave method queries unscoped.
 5. Store or update durable knowledge during work and before the final answer — verified findings, stated intent (goals, plans, direction the user voiced) stored as goals with user_stated provenance even while unsettled, and resolved diagnoses stored as skills when the cause will recur (an environment quirk, a config trap — not a typo). The test: could a fresh session recover this from memory alone? Never store raw transcripts, secrets, or transient debugging noise.
-6. Listed but not callable is not unavailable. A host may namespace the tools and withhold their schemas until asked, so load the schema for the name you need — namespaced it is `mcp__knowl__knowl_query` — and call it. Only when the tools are genuinely absent, stop and tell the user instead of silently bypassing Knowl.
+6. Listed but not callable is not unavailable. A host may namespace the tools and withhold their schemas until asked, so load the schema for the name your host lists and call it — where names are namespaced from the server key, `knowl_query` appears as `mcp__knowl__knowl_query`. Stop and tell the user only when the tools are genuinely absent, or when every call fails; never silently bypass Knowl.
 
 ### Lifecycle modes
 
--- a/KNOWL.md
+++ b/KNOWL.md
@@ -8,7 +8,7 @@
 3. Use a relevant active hit immediately. Inspect files only after a miss, conflict, stale/low-confidence memory, or explicit verification request.
 4. Query again before switching to a distinct subtask or project area, and before choosing how to build something new — existing tooling and pipelines are project knowledge, and in a linked workspace they often live in a sibling repo, so leave method queries unscoped.
 5. Store or update durable knowledge during work and before the final answer — verified findings, stated intent (goals, plans, direction the user voiced) stored as goals with user_stated provenance even while unsettled, and resolved diagnoses stored as skills when the cause will recur (an environment quirk, a config trap — not a typo). The test: could a fresh session recover this from memory alone? Never store raw transcripts, secrets, or transient debugging noise.
-6. Listed but not callable is not unavailable. A host may namespace the tools and withhold their schemas until asked, so load the schema for the name you need — namespaced it is `mcp__knowl__knowl_query` — and call it. Only when the tools are genuinely absent, stop and tell the user instead of silently bypassing Knowl.
+6. Listed but not callable is not unavailable. A host may namespace the tools and withhold their schemas until asked, so load the schema for the name your host lists and call it — where names are namespaced from the server key, `knowl_query` appears as `mcp__knowl__knowl_query`. Stop and tell the user only when the tools are genuinely absent, or when every call fails; never silently bypass Knowl.
 
 ### Lifecycle modes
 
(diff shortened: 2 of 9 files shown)


williamttruong:
## Measured the card, since the PR body left the ceiling question open

The claim in the decision record — that this is *"empirically checkable with the existing harness (`npm run benchmark:accuracy`)"* — **is wrong.** `grep -rln "OPERATIONAL_CARD|SERVER_INSTRUCTIONS|mcpServerInstructions|knowl-guidance" benchmarks/ scripts/` returns nothing. The retrieval suites drive `queryKnowledge()` directly; the ranker never sees the card. No harness in this repo can measure a card change.

So I used the instrument that already refuted the keyword cap: the real transcript archive. 348 unique main sessions, 143 days.

### Where the 2,000 came from

`git log -S` puts both the ceiling and its `20 *` multiplier in **`1a65701 feat: add canonical Knowl guidance renderers`** (2026-07-21) — the commit that *created* the renderers. Bare one-line message, no rationale. It is a requirement set before the thing it bounds existed.

For contrast, every neighbouring budget in `src/core/token-budget.ts` carries its derivation inline: `MAX_ITEM_CONTENT_CHARS` has a four-row cost table over 556 real items, `MAX_TITLE_CHARS` cites p50/p90/p99/longest, `MAX_AFFECTED_PATHS` cites percentiles over 710 items. The card ceiling is the only budget here without one.

Also worth noting: the three assertions at `tests/core/knowl-guidance.test.ts:132-137` are one assertion. For integer `n`, `ceil(x) <= n` iff `x <= n`, so `Math.ceil(len/4) <= 500` **is** `len <= 2000`, and `20 * Math.ceil(len/4) <= 10_000` is exactly that again. Only the first can fail; the `20×` buys no independent guarantee.

### What the card costs

| card | chars | tok/session | tok/month |
|---|---|---|---|
| claude | 1,833 | 459 | 20.2k |
| server | 1,884 | 471 | 20.8k |
| server + transcripts | 1,989 | 498 | 21.9k |
| the 2,000 ceiling | 2,000 | 500 | 22.0k |

210 sessions across 5 knowl-configured projects, 1.5 sessions/day. **29.5% never called a knowl tool at all** — they paid the card for nothing, and that is the common case, not an edge case.

**The 67-char line this PR could not fit: 17 tok/session, 0.75k tok/month.** For scale, `tools/list` in the **same handshake** is 22,394 chars = 5,599 tokens. The card is **8.9%** of what knowl already spends there.

### What the card buys

Sessions that read files at all, split on the card's creation date:

| | before (n=52) | after (n=114) |
|---|---|---|
| queried before first file read | 13.5% | **72.8%** |
| read files first | 26.9% | 20.2% |
| never queried at all | 59.6% | **7.0%** |

**This is observational, not an A/B.** The card, the hooks, `KNOWL.md` and the prompt reminder all shipped as one guidance system, so it measures the system, not the 2,000 characters in isolation. It says the guidance works; it cannot attribute the gain to the card alone.

### Where room exists, if you want it

Across 1,932 knowl calls: writes 62.5%, retrieval 33.5% — **96% of traffic**. Audit + skills + special together carry **0.7% of calls for ~450 characters**. The work-loop line is ~210 chars for **zero** observed calls — but zero is exactly what that rule asks for when hooks are active, so this data cannot separate "the prohibition works" from "dead weight here". Low usage is also not low value: a skill read once can be decisive.

### Two traps for anyone re-running this

1. The 2026-08-12 drive move left `d--*` archive directories that are **byte-identical** copies of `c--*` (53/53 filenames, same size, mtime, bytes). Counting both doubles every figure — my first run reported 422 sessions and 44k tok/month, both exactly 2× wrong.
2. Of 2,268 `.jsonl` files, only 349 are main sessions; **1,919 are subagent transcripts**, which never receive the MCP instructions block.

### What I am not doing

Not touching the card in this PR. The measurement says 67 characters cost 0.75k tok/month against a 21.9k baseline that is itself 8.9% of its own handshake — but spending them means compressing prose with a retrieval eval behind it, and that is your call. Harness coming as a separate PR so this one stays a guidance fix.

dat999zx:
Merged, with fixes pushed to the branch first (`902b266`). The diagnosis is right and the failure is real — this very session listed every knowl tool under "schemas are NOT loaded", so rule 6 was misfiring exactly as described. Four things needed finishing.

**1. Only the write half moved onto the constant.** Three readers kept the literal: `project-adapters.ts:48`, `project-adapters.ts:84`, `desktop-adapter.ts:20`. Mutating `KNOWL_MCP_SERVER_KEY` proved the split — every adapter reported `configure → configured`, then `verify → false` and `detect.configured → false`. That is `knowl init` writing a config `knowl doctor` can never see, and `doctor --fix` re-running init forever without converging. Your own framing — "nothing connected the two halves" — applied to the half left alone.

**2. The test guarding that could not fail.** `expect(KNOWL_NAMESPACED_TOOL_PREFIX).toBe(`mcp__${KNOWL_MCP_SERVER_KEY}__`)` restates the definition. Under the mutation the whole guidance file passed while rendering a name nothing registers; what actually caught it was `agent-adapters.test.ts` asserting the literal — which is precisely what a renamer would update in the same commit. It now reads the key back off the config that was written, plus a new adapter test driving `configure → detect → verify` across all five hosts. Verified load-bearing by re-breaking one read site: it fails.

**3. `mcp__…__` is Claude Code's convention, not a universal.** `installKnowlProjectGuidance` writes the same managed text into `KNOWL.md` and `AGENTS.md` whichever host was selected, and `GEMINI.md` includes it by reference — so stating it flatly hands Codex, Cursor, Gemini and Claude Desktop users an exact name that does not resolve. That is the failure class this PR set out to prevent, relocated. `host-hook.ts:138` already splits on the *last* `__` because it has met other shapes. The prose now names the form as what a namespacing host produces rather than as the name every host shows.

**4. The compact cards kept the misfiring sentence** — in the surface your own doctor comment names as one of only two carriers left on a deferred host. You were right that the ~230-char sentence does not fit. But `unavailable` → `absent` is **−5 chars** and removes the ambiguity outright, which is the actual bug. Cards are now **1,828 / 1,879 / 1,984**, headroom 16. Whether to buy room for the full instruction is still open — and #125 exists to price exactly that.

Also: rule 6 left a third case outside both branches — listed, schemas loaded, every call failing. Not deferred, not "genuinely absent", so an agent could read the stop instruction as not applying. Now covered. And the 12-line comment sat above `verifyInstructions`, an unrelated check, 28 lines from the messages it describes; moved, and it now records that the `degraded` branch is unreachable from doctor today (no adapter returns it — only `init-flow.ts` produces it, and that does not feed this call).

You were right to leave severity alone, and right that the card budget was your call to escalate rather than spend.

PR #125, opened 2026-08-16
eval(guidance): give the card ceiling the derivation its neighbours have

williamttruong:
Follow-up to #124, which hit the 2,000-character ceiling and could not justify spending against it. This measures the ceiling instead of arguing with it. **No behaviour changes — one script, one doc, one npm script.**

## The problem

The compact card is the one surface charged to **every session of every user**, and it was the only budget in this repo with no derivation. Its neighbours in `src/core/token-budget.ts` each carry one inline — `MAX_ITEM_CONTENT_CHARS` a four-row cost table over 556 real items, `MAX_TITLE_CHARS` p50/p90/p99, `MAX_AFFECTED_PATHS` percentiles over 710.

`git log -S` puts the ceiling and its `20 *` multiplier in **`1a65701`**, the commit that *created* the renderers, with a bare one-line message. It is a requirement set before the thing it bounds had any measured cost.

Also: those three assertions at `tests/core/knowl-guidance.test.ts:132-137` are **one** assertion. For integer `n`, `ceil(x) <= n` iff `x <= n`, so `Math.ceil(len/4) <= 500` *is* `len <= 2000`, and the `20 *` line is that again. Only the first can fail.

**The accuracy harness cannot supply a derivation.** `benchmark:accuracy` and the retrieval suites drive `queryKnowledge()` directly — the ranker never sees the card. Grepping `benchmarks/` and `scripts/` for the card constants returned nothing. (This contradicts the closing line of the "card carries policy, not an inventory" decision record, which offered that harness as the check.)

## What this adds

- **`scripts/measure-card-cost.ts`** (`npm run measure:card`) — reads the host transcript archive, prints cost and observed use. Read-only, no network. `--archive` / `--since`.
- **`docs/evals/guidance-card-cost.md`** — the writeup, in the shape of the other eval docs.

Card sizes are **rendered from `knowl-guidance.ts`**, never hard-coded, so the report cannot drift from the card. A wrong config shape now throws rather than silently reporting the transcript variant as the same size as the server card — which it did on my first run, overstating headroom tenfold (116 vs 11).

## Results

210 sessions, 5 projects, 143 days. **29.5% never called a knowl tool** — they paid the card for nothing, and that is the common case.

| card | chars | tok/session | tok/month |
| --- | --- | --- | --- |
| claude | 1,833 | 459 | 20.2k |
| server | 1,884 | 471 | 20.8k |
| **server + transcripts** | **1,989** | **498** | **21.9k** |

Headroom **11 chars**. 100 more characters cost **1.10k tok/month**. `tools/list` in the *same handshake* is 5,599 tokens — **the card is 8.9% of what we already spend beside it.**

Only 2,000 is a decision; 1,833/1,884/1,989 are one card in three configurations. Worth noting 1,833 and 1,884 are pinned exactly while **1,989 is only bounded** — the variant that binds is the one no test pins.

**What it buys**, split on the card's creation date, sessions that read files:

| | before (n=52) | after (n=114) |
| --- | --- | --- |
| queried before first file read | 13.5% | **72.8%** |
| never queried at all | 59.6% | **7.0%** |

**Observational, not an A/B** — the card, hooks and `KNOWL.md` shipped as one system. It says the system works; it cannot attribute the gain to the 2,000 characters alone. The doc states this where the number appears.

**Where room exists:** writes 62.5% + retrieval 33.5% = 96% of 1,933 calls. Audit + skills + special = **0.7% of calls for ~450 chars**. The work-loop line reads zero — but zero is what its rule asks for when hooks are active, so this cannot separate "the prohibition works" from "irrelevant here", and low volume isn't low value.

## Two traps documented for anyone re-running it

1. The drive move left `d--*` archive trees **byte-identical** to `c--*` (53/53 filenames, same bytes). Counting both doubles everything — my first run said 422 sessions / 44k tok/month, exactly 2× wrong.
2. Of 2,268 `.jsonl` files only **349 are main sessions**; 1,919 are subagent transcripts, which never receive the instructions block.

## What this does not settle

Whether to spend the room. Buying it means compressing prose with a retrieval eval behind it, and this measures cost and observed use, not what compression does to routing quality — pair it with `benchmark:accuracy` before shipping a trim.

## Verification

`tsc --noEmit` clean · `eslint .` clean · `docs:check` current · **337 test files, 3,098 tests, 0 failures**

### DIFF
--- a/docs/evals/guidance-card-cost.md
+++ b/docs/evals/guidance-card-cost.md
@@ -1,8 +1,14 @@
 # The guidance card: what it costs, what it buys, and where its ceiling came from
 
-Run 2026-08-16 against this machine's Claude Code transcript archive. Harness:
-`npx tsx scripts/measure-card-cost.ts` (`npm run measure:card`). Nothing is written and no network
-is touched; card sizes are rendered from `src/core/knowl-guidance.ts`, never hard-coded.
+Run 2026-08-18 against this machine's Claude Code transcript archive, at `5957bb7`. Harness:
+`npm run measure:card` (`scripts/measure-card-cost.ts`). Nothing is written and no network is
+touched.
+
+**Every number below is rendered or counted by the script.** Card sizes come from
+`src/core/knowl-guidance.ts` and the `tools/list` payload they are compared against comes from
+`knowlToolDefinitions`. That is not a stylistic preference: the first version of this document
+hard-coded the `tools/list` comparison, and it was already wrong by a third — 22,394 characters
+against a real 34,507 — which inverted the conclusion it was there to support.
 
 **Why this exists.** The compact card is the one surface charged to every session of every user,
 and it was the only budget in this repository with no derivation. Its neighbours in
@@ -19,7 +25,7 @@ cites percentiles over 710. The 2,000-character ceiling had none.
 renderers. The message is one bare line with no rationale. The number therefore predates the thing
 it bounds having any measured cost: it is a requirement, not a finding.
 
-It is also asserted three times where once would do. At `tests/core/knowl-guidance.test.ts:132-137`:
+It is also asserted three times where once would do. In `tests/core/knowl-guidance.test.ts`:
 
 ```js
 card.length < 2_000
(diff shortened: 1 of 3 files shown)


dat999zx:
Reviewed, fixed and rebased onto `main` (`5f0e148`). The instrument is good — portable, read-only, lint and type clean, and the card sizes it renders are exact. The analysis built on it had four defects, and one of them inverted the argument the document exists to make.

**1. `tools/list` was hard-coded and stale.** The doc says 22,394 chars = 5,599 tokens, "the card is 8.9%". Measured from `knowlToolDefinitions` today: **30 tools, 34,507 chars, ~8,627 tok — the card is 5.7%.** That figure is the entire scale argument, in a PR whose selling point is *"card sizes are rendered from `knowl-guidance.ts`, never hard-coded, so the report cannot drift."* The script measured the card from source and then compared it to a number typed by hand. It now renders both.

**2. The cost rate divided by days the card did not exist.** 143 days, of which **117 predate the 2026-07-21 creation date** in your own §0. The window now starts at the card by default, with `--all` to widen and a line in the output saying which is in force. Same fix retires the "29.5% paid for nothing" figure, which was counting pre-card sessions in the overhead bucket.

**3. The group table did not sum.** `total++` ran for every knowl call; the row assignment sat behind `if (hit)`. Your own §4 sums to 97.8%. And no pattern matched the transcript tools — so the table asking *where is there room* was structurally blind to `TRANSCRIPT_ROUTE_LINE`, the **+105 characters that make the card binding**. Added a `transcripts` group and an explicit `other (unrouted)` row. `other` turns out to be `knowl_cloud`, called 13 times and named nowhere in the routing table — worth its own look.

**4. Dedup kept the wrong copy, silently.** It keeps whichever directory `readdirSync().sort()` returns first and verifies nothing. ASCII sort puts `C--x` before `c--x`, so after the drive move it reliably kept the **stale pre-move tree** and threw away the live one with every call recorded since. Now keeps the newest by mtime and reports copies that differ in size — "byte-identical" is an assumption about a disk, not a fact.

Also: `--since` with a bad date said "No sessions matched" instead of naming the bad date; `arg()` accepted `--archive --since X` as `ARCHIVE=--since`; the config guard threw above `main()` so it printed a stack trace instead of its own message; the transcript config is built from `DEFAULT_CONFIG` now, so a renamed field fails to compile; and the skipped-subagent count is printed rather than asserted in prose.

**Numbers after the rebase** (this archive, 186 sessions / 28 days, and note #124 changed the cards):

| card | chars | tok/month |
| --- | --- | --- |
| claude | 1,828 | 90.2k |
| server | 1,879 | 92.7k |
| **server + transcripts** | **1,984** | **97.9k** |

Headroom 16. 100 chars = 4.93k tok/month. Card is 5.7% of `tools/list`.

**§3 now says it cannot answer.** This archive starts six days before the card shipped, so the before column is n=3. I did not restate your 13.5% → 72.8% as if I had measured it — that came from a longer archive and this run cannot see it. If you re-run `npm run measure:card -- --all` on yours, that section is yours to fill in properly.

Your two framing calls were both right: the redundant-assertion argument is mathematically correct, and `benchmark:accuracy` genuinely cannot see the card. The `claude` card row is now marked priced-for-comparison-only, since nothing in `src/` actually delivers `KNOWL_CLAUDE_OPERATIONAL_CARD` — the `SessionStart` wiring a 2026-07 design doc describes was never built.

PR #133, opened 2026-08-19
fix(store): a claim can no longer be retired by its own negation

williamttruong:
## The defect

`resolveDuplicate` supersedes whenever `sameSubjectTitle` holds, and that is a token-**subset** test. None of `not`, `no`, `never`, `longer` is in `DUPLICATE_STOP_WORDS`, so an affirmative title is a strict subset of its own negation and the subset test fires.

Found by running the exported function, not by reading it:

```
coexist    | new="Vector dim is 768"               vs held="Vector dim is 1024"
SUPERSEDE  | new="Reranker is the right call"      vs held="Reranker is not the right call"
SUPERSEDE  | new="Reranker is not the right call"  vs held="Reranker is the right call"
coexist    | new="Retrieval latency p50 is 208ms"  vs held="Retrieval latency p50 is 428ms"
SUPERSEDE  | new="Push gate blocks default branch" vs held="Push gate no longer blocks default branch"
SUPERSEDE  | new="Cloud send works"                vs held="Cloud send works end to end in production"
```

The survivor then asserts the opposite of what was retired, silently. **Row 5 is not hypothetical** — it is this project's own 2026-08-13 push-gate reversal. A title shape written twice across that reversal retires its own predecessor, in whichever order the two writes happen to arrive.

## The fix

Two guards on the supersede branch, both clamping to `coexist` rather than refusing. **Neither can lose a write:** the incoming atom is still inserted, the predecessor stays active, and the pair is reported through the `nearDuplicate` channel that already hands the caller the exact retire call. An explicit `supersedes` is checked first and is never second-guessed.

1. **Polarity** — if everything the longer title adds is a polarity token, the two titles are the same claim asserted both ways, not "same subject, new information".

2. **Grounding** — an atom that does not claim first-hand provenance no longer retires one that does ("inferred evidence must not overturn an explicit fact"). One-directional and narrow: it fires only where the *held* atom claims `observed`/`user_stated` and the incoming one does not, so the ~79% of writes that leave provenance unset are unaffected unless aimed at one of the ~20% that made a claim. That asymmetry is the same incentive `fuse.ts` already chose on the ranking side — declaring provenance is what earns full credit, and here it earns the right to retire a grounded claim.

## What I deliberately did NOT port

The published prior art for this guard (muninndb's `dedup_separation.go`) is built around **numbers** — "$99" merging with "$149". **That hole does not exist here.** `duplicateTokens` splits on `[^a-z0-9_]+`, so digits survive as tokens, `768` is absent from a title saying `1024`, the subset test already fails, and the two atoms already coexist — rows 1 and 4 above. Porting its regex bank would have been dead weight while leaving the real hole open. Only polarity words are short, common, and absent from the stop list.

Ambiguous stems are also left out — `can` ("Can the gate block…" is a real title), plus `won` and `don`. A missed negation costs what today already costs; a false positive costs a supersede the caller then makes deliberately.

## Verification

- **16 new tests.** Disabling the polarity guard fails exactly the two load-bearing ones — including the end-to-end case, which reads `superseded` without the guard and `active` with it. That is the regression pin.
- `tsc --noEmit` clean.
- Full suite **3017 passed**. 21 files fail, and **the identical 21 fail on a pristine `upstream/main` worktree** — Windows CLI-subprocess tests plus `tests/store/retention.test.ts`. Baseline, untouched by this change; happy to attach both JSON runs.

Not for self-merge — over to you.

### DIFF
--- a/src/store/knowledge-writer.ts
+++ b/src/store/knowledge-writer.ts
@@ -36,6 +36,39 @@ const DUPLICATE_STOP_WORDS = new Set([
   'it', 'of', 'on', 'or', 'our', 'the', 'this', 'to', 'use', 'uses', 'using', 'what', 'with',
 ]);
 
+/**
+ * Tokens that flip a claim rather than extend it.
+ *
+ * THE FAILURE THIS EXISTS FOR. `sameSubjectTitle` is a token-SUBSET test, and none of these
+ * words is a stop word, so an affirmative title is a strict subset of its own negation and the
+ * subset test fires. Measured by running the real function, 2026-08-19:
+ *
+ *     "Push gate blocks default branch"  SUPERSEDES  "Push gate no longer blocks default branch"
+ *     "Reranker is the right call"       SUPERSEDES  "Reranker is not the right call"
+ *
+ * The survivor then asserts the opposite of what was retired, silently. The first pair is this
+ * project's own 2026-08-13 push-gate reversal: a title shape that, written twice across the
+ * reversal, retires its own predecessor in whichever order the two writes happen to arrive.
+ *
+ * WHY NOT THE NUMERIC CASE TOO. The published prior art for this guard (muninndb's
+ * `dedup_separation.go`) is built around numbers -- "$99" merging with "$149". That hole does not
+ * exist here and the check would be dead weight: `duplicateTokens` splits on `[^a-z0-9_]+`, so
+ * digits survive as tokens, `768` is absent from a title saying `1024`, the subset test already
+ * fails, and the two atoms already coexist. Verified alongside the two cases above. Only polarity
+ * is unguarded, because only polarity words are short, common, and absent from the stop list.
+ *
+ * DELIBERATELY NARROW. Ambiguous stems are left out -- `can` ("Can the gate block" is a real
+ * title), `won`, `don`. A missed negation costs what today already costs; a false positive here
+ * costs a supersede that should have happened, which the caller then has to make deliberately.
+ * Contraction apostrophes are split by the tokenizer, so `isn't` arrives as `isn` plus a
+ * one-character token that the length filter drops -- hence the bare stems below.
+ */
+const POLARITY_TOKENS = new Set([
+  'not', 'no', 'never', 'none', 'nor', 'without', 'cannot', 'longer', 'unable',
+  'isn', 'aren', 'wasn', 'weren', 'doesn', 'didn', 'hasn', 'haven', 'hadn',
+  'couldn', 'shouldn', 'wouldn',
+]);
+
 export interface StoreKnowledgeInput {
   category: KnowledgeCategory;
   title: string;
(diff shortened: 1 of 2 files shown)


dat999zx:
Reviewed and verified by running it rather than reading it. **The polarity guard is right and I've kept it. The grounding guard I've removed** — pushed to this branch as `34e05f5`.

## The polarity guard stays

Real bug, correct fix, and load-bearing: disabling it fails 3 tests including your end-to-end pin.

I replayed the predicate over every real supersession in this repo's store — `knowledge_items` self-joined on `superseded_by_id`, 101 pairs — and over all same-category active pairs:

- **0 of 101** supersessions would have been blocked
- **0** active polarity twins exist

So the cost is measurably zero. It's free insurance, which is the best case for a guard like this. One correction to the body though: "Row 5 is not hypothetical — it is this project's own 2026-08-13 push-gate reversal" describes a title *shape* that would misbehave, not a supersession that occurred. The store says it never fired. Doesn't make the fix wrong, but it shouldn't be quoted as damage taken.

## The grounding guard I removed

Same replay, and this one does fire — **3 of 101**. All three are legitimate corrections:

```
HELD(observed): A helper reachable from a tool call must swap databases with withDbPath...
NEW (unset)   : Only withDbPath is safe inside an MCP request; initDbPath and closeDb own...

HELD(observed): Review blockers at 626799c: consent, ledger state, confirmation snapshot...
NEW (unset)   : 5.0 spec revised at 6a8eae3: all eight review findings resolved...

HELD(observed): Guidance card sizes on main are 1,833 / 1,884 / 1,989 with 11 chars headroom...
NEW (unset)   : The guidance card costs ~98k tok/month ... and four measurement defects had
                understated or misdirected every earlier figure
```

The third is the one that decided it. That atom's entire purpose is to say the held measurement was defective — and the guard keeps the defective figure active. Small blast radius, 100% error rate inside it.

Three things behind that:

**The premise doesn't hold on this corpus.** Unset provenance isn't a weak claim, it's what most writes look like: 680 unset / 251 `observed` / 6 `user_stated` across 937 active items. The carefully verified atoms sit in the unset majority as often as anywhere else, so the absence of a label carries no signal about the write.

**The ranking symmetry doesn't carry.** It's `agent-query.ts:150`, not `fuse.ts` — and `UNCLAIMED_PROVENANCE_PRIOR` is `0.98`. A 2% nudge in ordering and a refusal to supersede aren't the same judgement at different volumes. Its own comment says it's meant to bite "only where items differ, which is the only place it should"; a supersede/coexist flip is a much larger lever than that was scoped for.

**"Clamping to coexist can't lose a write" is true of the write, not of the store.** `nearDuplicate` fires once, to whoever reads the write result — a hook or the commit extractor has nobody to act on it. And nothing re-flags the pair afterwards: `listActiveConflictKeys` reads only `conflictKey`/`conflictExclusive`, set on **3 of 937** active items. So a contradiction parked by that guard is invisible from then on, which is the same end state the polarity guard exists to prevent.

The measurement is now recorded at the call site and pinned by two tests, so the idea doesn't come back without new evidence. If you can show a corpus where an unclaimed write retiring an `observed` one caused real damage, I'll take it back — this is a measurement, not a principle.

## Verification

All six CI gates pass: `typecheck`, `lint`, `build`, `docs:check`, `typecheck:bench`, `check:lockfile`. `tests/store` + `tests/core` + `tests/mcp`: **160 files, 1504 passed, 0 failed**.

One note on your baseline claim — `tests/store/retention.test.ts` passed in all four of my runs here, on both branches. Worth re-checking what made it fail for you; it may be a local-state thing rather than a baseline failure.

Thanks for this one. The subset-test hole is a good catch and the way you found it — running the exported function over a table of real title shapes instead of reasoning about it — is exactly right.

PR #134, opened 2026-08-19
feat(core): keep both ends of a truncated atom body, not just the head

williamttruong:
## The problem

Atom bodies in this project put the verdict last — `CONSEQUENCE`, `THE FIX`, `WHAT NOT TO BUILD`, `THE LIMIT OF THIS EVIDENCE`. `truncateText` is a head slice, so it systematically delivers the setup and drops the finding.

That would be recoverable if the `truncated` flag were acted on. Measured over the session archive, it mostly is not: **43.3% of delivered items arrive truncated, and only 2.2% of them are ever chased by an id-fetch.** The withheld tail isn't being declined — it's going unread.

That same measurement recorded its own limit, and it is exactly this: it *"cannot distinguish 'the tail was not needed' from 'the agent answered from a partial fact and never discovered it was wrong'."* A head slice is what makes the second case invisible. (Fittingly, the atom recording that finding is itself long enough to be truncated, and `THE LIMIT OF THIS EVIDENCE` is in the part that gets cut.)

## The fix

`truncateMiddle` — 60/40 head/tail around a fixed marker. The head keeps the larger share because it decides whether the atom is relevant at all; the tail exists because that is where the atom says what it concluded.

**This is not a ceiling change, and that is the point.** The corpus sweep that settled `MAX_ITEM_CONTENT_CHARS` at 2,000 showed that raising it to 4,000 buys content declined 97.8% of the time at +399 tok/query. This spends **nothing** — the budget is identical to the character, and only *which* characters survive changes. It's orthogonal to that decision rather than a re-litigation of it.

Exact by construction (`head + marker + tail === maxChars` whenever there's room for the marker), so `token-budget.test.ts`'s existing length assertion holds unchanged.

## Scope

Applied to the three surfaces that hand a stored body to an agent:

- `compactKnowledgeItem` — every `knowl_query` result
- `knowl://category/{name}` — truncate-then-contain order preserved, so the fence length is still picked over the text that actually ships
- `task_start`'s `relevantMemory`

Titles, tags and `affectedPaths` deliberately keep the head slice: a path with its middle removed is not a shorter path, it's an unusable one, and none of those fields has a conclusion at the end worth preserving.

A byte count in the marker was considered and rejected, following the `truncated` flag's own stated reasoning — the caller is deciding "do I need the rest of this", and a length answers a question nobody asked. It also keeps the arithmetic exact.

## Verification

- **13 new tests**, including a pin that a trailing caveat survives where a head slice provably drops it, and one asserting paths are *not* middle-elided.
- Existing `token-budget.test.ts` passes unchanged.
- `npm run typecheck` clean.
- Suite excluding `tests/cli`: **2581 passed, 1 failed** — `tests/store/retention.test.ts`, which fails identically on a pristine `upstream/main` worktree. (`tests/cli/**` has 20 pre-existing failures on this machine, also identical at baseline.)

## Not included

The branch name mentions atom markers (`<knowl_atom id rev>` wrappers for dedup and mid-session invalidation). I've left that out — it only pays off alongside a consumer that reads the markers back, and it touches the MCP response surface whose block count is pinned by three tests. Better as its own PR.

Not for self-merge — over to you.

### DIFF
--- a/src/core/token-budget.ts
+++ b/src/core/token-budget.ts
@@ -157,6 +157,49 @@ export function truncateText(value: string, maxChars: number, marker = ''): stri
   return `${value.slice(0, maxChars - marker.length)}${marker}`;
 }
 
+/**
+ * What replaces the elided middle. Fixed rather than counted, following the `truncated` flag's
+ * own reasoning: the caller is deciding "do I need the rest of this", and a byte count answers a
+ * question nobody asked. A fixed marker also keeps the arithmetic below exact.
+ */
+const MIDDLE_ELISION = '\n\n[... elided ...]\n\n';
+
+/**
+ * How much of the surviving budget goes to the opening. The head decides whether the atom is
+ * relevant at all, so it keeps the larger share; the tail is there because that is where the
+ * atom says what it concluded.
+ */
+const HEAD_SHARE = 0.6;
+
+/**
+ * Keep both ends of an over-long body and drop the middle.
+ *
+ * WHY THIS EXISTS, AND WHY IT IS NOT A CEILING CHANGE. Atom bodies in this project are written
+ * with the verdict last -- `CONSEQUENCE`, `THE FIX`, `WHAT NOT TO BUILD`, `THE LIMIT OF THIS
+ * EVIDENCE` -- so a head slice systematically delivers the setup and drops the finding. That
+ * would be recoverable if the flag were acted on, but it mostly is not: measured over the
+ * session archive, 43.3% of delivered items arrive truncated and only 2.2% of them are ever
+ * chased by an id-fetch. The withheld tail is not being declined, it is going unread.
+ *
+ * The corpus measurement that settled `MAX_ITEM_CONTENT_CHARS` at 2,000 also showed that raising
+ * it to 4,000 buys content declined 97.8% of the time at +399 tok/query. This spends nothing:
+ * the budget is identical to the character, and only the choice of WHICH characters changes. It
+ * is therefore orthogonal to that decision rather than a re-litigation of it.
+ *
+ * Exact by construction: `head + marker + tail === maxChars` whenever there is room for the
+ * marker at all, which is what keeps the existing length assertion true.
+ */
+export function truncateMiddle(value: string, maxChars: number): string {
+  if (value.length <= maxChars) return value;
+  // With no room for the marker there is no middle to speak of, and a caller asking for fewer
+  // characters than the marker wants a prefix, not two fragments of one.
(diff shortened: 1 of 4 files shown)


dat999zx:
Reviewed and verified. **Merging as-is** — no changes needed.

The argument that sold it is that this is not a ceiling change. The budget is identical to the character; only which characters survive moves. That keeps it orthogonal to the 3.1.0 decision rather than a re-litigation of it, and `head + marker + tail === maxChars` means the existing length assertion holds without being touched.

## What I checked beyond the tests

**The escape hatch is still real.** `truncateMiddle` reaching `compactKnowledgeItem` would be a problem if the id-fetch shared that path — it doesn't. `tools.ts:733` builds its own object with `content: item.content` raw, so "call again with `id` to read it in full" stays literally true.

**No benchmark moves.** `mab-bridge.ts:179` returns `item.content` straight off `queryKnowledgeForAgent` and never touches `compactKnowledgeItem`, so this is purely an MCP delivery change. LongMemEval and MAB scores are untouched.

**On real atoms it does what it says.** Against the three longest active items in this repo's store, comparing the last 200 chars delivered today vs with this change:

```
head slice : "...Pushed to the contributor's fork via `maintainerCanModify` (`git push
              https://github.com/William-Sommers/knowl.git <local>:<their-branch>`)"
middle-elide: "...raised in #116's own body and not fixed there.

               **Unreleased now holds three sections** and no release has been cut for them."
```

```
head slice : "...**A canary, not an attestation.** Workspace stores "
middle-elide: "...\"Drop the server embedder entirely (removes 4 unfixable sharp CVEs, but
               leaves preset changes and web-UI writes with no owner)\"]"
```

Two of three end on the actual conclusion instead of mid-sentence. The third is a wash — the tail happened to be a serialized `alternatives` array rather than prose. Net positive at zero cost, which is the right trade.

**Corpus check on your 43.3% figure.** 19.4% of the 937 active atoms in this store exceed 2,000 chars. Not a contradiction — retrieval skews toward longer, richer atoms, so a higher share of *deliveries* than of the *corpus* is what you'd expect. Consistent.

The call to leave titles, tags and `affectedPaths` on the head slice is right, and worth having stated explicitly in the comment.

## Verification

All six CI gates pass: `typecheck`, `lint`, `build`, `docs:check`, `typecheck:bench`, `check:lockfile`. `tests/store` + `tests/core` + `tests/mcp`: **160 files, 1501 passed, 0 failed**.

One note — `tests/store/retention.test.ts` passed in every run here. Whatever made it fail on your machine looks local rather than a baseline failure; worth a look before you lean on it as a baseline again.

Good call leaving the atom markers out. Agreed they need a consumer first.

Issue #145, opened 2026-08-20
Evidence: two locator formats in one table, and three dead-or-latent paths around isEvidenceStale


williamttruong:
Found while auditing staleness detection. None of this is urgent — every path below is currently unreachable — but each becomes live the moment somebody uses the feature, and two of them are cheap to close now and expensive later.

Measurements are from this repo's own store (`knowl-cloud/.knowl/knowl.db`, 2026-08-20).

## 1. Two locator formats coexist in `evidence`

`createEvidence` normalizes through `normalizeLocator`, which returns a scheme-prefixed value: `file://src/store/read-set.ts`.

Every row in the live table is a **bare path**:

```
file evidence rows       : 588
  ...with file:// prefix : 0
  ...with a contentHash  : 0
symbol evidence rows     : 0
```

So something writes `evidence` while bypassing `createEvidence`, and the two halves of the system disagree about what a locator looks like.

**This is latent, not live, and the reason is worth stating precisely.** `isEvidenceStale` short-circuits on `if (evidence.type !== 'file' || !evidence.contentHash) return false;`. With zero rows carrying a hash, the `readFile` below it never executes.

The moment `contentHash` starts being populated, whichever format is wrong breaks:

- bare path (`web/lib/api.ts`) → `path.resolve(root, locator)` is correct, works
- prefixed (`file://web/lib/api.ts`) → resolves to `<root>/file:/web/lib/api.ts`, which cannot exist → `readFile` throws → the `catch` returns `true` → **every file evidence reports stale, permanently**

Since `createEvidence` is the normalizing path and it produces the *prefixed* form, the format that breaks is the one the intended API writes.

Two questions, both yours: which format is canonical, and what is writing rows that skip `createEvidence`?

## 2. `command` and `url` are in `EvidenceType` and reach nothing

```ts
export type EvidenceType = 'file' | 'symbol' | 'commit' | 'test' | 'command' | 'url' | 'user' | 'agent';
```

`isEvidenceStale` handles `symbol` and `file`; everything else returns `false`. So `command` and `url` are storable, syncable, accept free-text locators and arbitrary metadata JSON, and are checked by nothing.

**The live store has zero rows of either.** Deleting them is free today.

The reason to consider it now rather than later: they are exactly what a future implementer reaches for when adding any kind of verification, and a stored free-text `command` that something later executes is a different security posture than this codebase currently has. Removing an unused member of an exported union is a breaking type change, so it is your call — I am flagging the window, not arguing it should close.

[...]

PR #153, opened 2026-08-20
fix(evidence): a file locator could name a path outside the repository

williamttruong:
Closes the containment half of #145 — items (1)–(3). Leaves (4) to #152 and the sync/import trust-boundary question to its own issue, as agreed.

## What was wrong

`createEvidence` normalized every locator through a private `normalizeLocator` that did `trim().replace(/\/g, '/')` and no segment check. `../../../Windows/win.ini` stored verbatim, and `isEvidenceStale` hands the stored string to `path.resolve(projectRoot, ...)` before `readFile`.

**The `contentHash` short-circuit is what makes this latent — not the paths being safe.** No row in either store carries a hash, so the read never happens. I've written the commit message to say that rather than "the paths are safe", since that distinction is the one §1 of the issue originally lost.

Your `sync-apply.ts:147-163` find is why this shouldn't wait: the window doesn't close locally, and nothing in the commit that closes it will look like it touched security.

## Two deviations from the agreed spec, both deliberate

I flagged these in the issue before building; restating so the diff isn't surprising.

**1. Shares the segment check, not `normalizeLocator` wholesale.** `normalizeFilePath` requires an extension in the basename — deliberately, per your docstring, to reject directories, losing `Makefile`, `LICENSE` and `.gitignore` as an acceptable cost for a tier "allowed to be incomplete". Evidence is a deliberate citation, not that tier, so wholesale delegation would start refusing writes that work today. New exported `containedRepoPath` is the containment half without the bias. One rule about `..`, one place — which is the property you actually wanted.

**2. Canonical form is per type.** Bare path for `file` as you said, but `symbol` must keep its scheme: `symbol://path#name` *is* `code_symbols.locator`, the primary key `resolveSymbolEvidence` looks up (`evidence-repository.ts:151`). Stripping it uniformly would make every symbol evidence miss its exact lookup. Zero symbol rows today, so it was a trap for the first writer rather than a live break. Non-path types are untouched — a commit SHA is an identifier, and validating it as a path would reject the 604 agent and 168 commit rows you measured.

**Also caught while writing the tests:** a drive letter clears the segment check. `C:/Windows/win.ini` has no empty, `.` or `..` segment, and `path.resolve` abandons the root for it — your own §1 table shows it escaping. Refused on every platform, since a repo-relative path begins with one nowhere.

## Guarded on both sides

Write-side alone is not sufficient, by your own finding: `cloud/sync-apply.ts` and `store/portability.ts` both insert without passing through any normalizer. So `isEvidenceStale` re-checks containment at the point of use. It refuses outright rather than resolving and letting `readFile` fail — that keeps the return value independent of anything outside the root, so it isn't a file-existence oracle even for one bit.

## Two calls that are yours

**Rejection throws.** A path that escapes the root isn't a locator, and silently coercing produces a row that looks fine and cites nothing. But it turns a call that always succeeded into one that can reject — one line to make it store-and-warn if you'd rather.

**`file://` is canonicalized, not rejected.** You asked for the test to assert rejection. I stripped-and-accepted instead: the prefix is a legitimate spelling of the same file, and canonicalizing is what actually ends "two formats in one column". The principle behind your request still holds — the test pins it as *intended* behaviour rather than as the accidental containment it is today. Happy to flip it if you disagree; it's a two-line change and the test says exactly which behaviour is being pinned.

## Verification

- New tests: both escape shapes plus a leading slash; `file://` canonicalization; an extensionless file accepted; non-path types opaque with the symbol scheme intact; and a row inserted directly via SQL to reproduce the sync shape and confirm the read-side guard catches what the write-side cannot see.
- `tests/store/evidence-repository.test.ts` + `tests/store/read-set.test.ts`: **32/32 green**.
- `tsc --noEmit` clean.
- Full suite: **21 files / 80 tests fail identically on unmodified `upstream/main`** — I ran the baseline before claiming this. Same 21 files, same 80 tests, and this branch is +5 passing.

### DIFF
--- a/src/store/evidence-repository.ts
+++ b/src/store/evidence-repository.ts
@@ -3,9 +3,9 @@ import fs from 'node:fs/promises';
 import path from 'node:path';
 import { Evidence, EvidenceInput, EvidenceRelationship, EvidenceType, KnowledgeEvidence, KnowledgeItem } from '../core/types.js';
 import { validateKnowledgeWrite } from '../core/knowledge-validation.js';
-import { getClient } from './database.js';
+import { getClient, getProjectRoot } from './database.js';
 import { getKnowledgeItem } from './repository.js';
-import { containedRepoPath, normalizeLocator as normalizeReadSetLocator } from './read-set.js';
+import { containedRepoPath, normalizeLocator as normalizeReadSetLocator, repoRelativePath } from './read-set.js';
 
 const MAX_EXCERPT_LENGTH = 2_000;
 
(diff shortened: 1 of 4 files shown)


dat999zx:
Reviewed, verified, and answered both calls you left me. Pushed one commit on top (`a138ee1`) rather than asking you to turn it around.

**The diagnosis is right and the read-side guard is the fix.** `createEvidence`'s private `normalizeLocator` had no segment check, and you are right that the `contentHash` short-circuit is what makes it latent rather than the paths being contained. Writing the commit message that way was the correct call — that distinction is the one §1 of #145 lost, and it is why this could not wait on `sync-apply.ts`.

**Rejection throws — no, store and neutralize.** You called this one as mine, and the deciding evidence was in the store rather than the code. This repo's `.knowl/knowl.db` has 1237 evidence rows, and three of them carry a locator your guard refuses:

```
file :: D:/coding/DuckPrep-server/KNOWL.md
file :: D:/coding/DuckPrep-server/AGENTS.md
file :: src/transcripts/
```

Two absolute paths naming a sibling checkout, and one directory refused only because a trailing slash reads as an empty segment — it escapes nothing. All three were written by agents doing ordinary work.

That matters more than a lost citation, because `attachEvidenceToKnowledge` turns every `affectedPaths` entry into a `file` locator and runs inside the write transaction in `storeKnowledgeItemDeduped`. A throw rolls the item back with it, so one bad path loses the whole atom, not the evidence.

And it defends nothing, by your own argument: `cloud/sync-apply.ts` and `store/portability.ts` insert without passing through any normalizer, which is exactly why `isEvidenceStale` has to check containment itself. Once it does, refusing at the write end is pure cost. Your containment assertion is unchanged and still the test that carries the fix.

I went one step further than store-and-warn: a path that names a file *inside* the repo by another spelling is now recovered rather than kept verbatim, reusing `repoRelativePath`. `D:/repo/src/a.ts` becomes `src/a.ts`, `src/transcripts/` loses its slash. Two spellings of one file in one column is a staleness check that never fires, and per `read-set.ts`'s own docstring that failure is silent by construction.

Order is resolve-then-check, deliberately: `containedRepoPath` decides whether the result is repo-relative, because your drive-letter rule holds on every platform while `path.resolve` only abandons the root for `C:/...` on Windows.

**`file://` canonicalized rather than rejected — agreed, keep it.** You were right and my request in the issue was wrong. Canonicalizing is what actually ends two formats in one column; asserting rejection would have pinned the accident.

**Deviation 1 (`containedRepoPath` rather than delegating to `normalizeFilePath`) — right call.** The extension requirement is a recall trade for a tier allowed to be incomplete, and evidence is not that tier. **Deviation 2 (per-type canonical form) — right call, and the better catch.** Stripping the scheme would have made every symbol row miss `code_symbols.locator`. Zero rows today, so you found a trap for the first writer rather than a live break.

**One thing your branch introduced.** `evidenceKey` in `knowledge-writer.ts:340` documents itself as "matched the way `createEvidence` stores it" and still did `trim` plus separator folding — the form `createEvidence` stopped using here. A re-store citing `file://src/a.ts` keyed differently from the stored `src/a.ts`, so a write that added nothing read as new information and retired the record it duplicated. It calls the shared normalizer now. Same class as #144, which is a fair thing to catch on a branch and an unfair one to catch after it lands.

**On the baseline.** The description reports 21 files / 80 tests failing identically on unmodified `main`. They do not fail on a built tree — that measurement was against a stale `dist/`, since twelve test files execute `dist/index.js`. Built first, `main` is green and so is this branch: 344/344 files, 3216 passed, 5 skipped, `tsc --noEmit` clean, `git diff --check` clean. Worth knowing before it becomes a habit; running `npm test` without `npm run build` validates the previous build, which reads as a green gate against code that no longer exists.

Merging.

PR #180, opened 2026-08-24
feat(conflicts): a reversal stored under an unrelated title gets a voice, and knowl conflicts stops missing its own pairs

williamttruong:
## The silence this closes

Two MCP store calls in a fresh store, verified against current `main` (7dda6a4) through the exact JSON-RPC path an agent uses:

1. `knowl_store` decision **"Database choice: Postgres for everything"**
2. `knowl_store` decision **"We are moving persistence to SQLite"**, whose content literally says *"The Postgres-for-everything plan is abandoned."*

Before this change: the second write returns bare success, `knowl_query "database persistence choice"` returns **both decisions as fresh**, and `knowl conflicts` returns **`[]`** — the command whose description promises "knowledge items that contradict each other" read only `conflictKey`/`conflictExclusive`, declared on 3 of 937 active items in the store that motivated this. The titles share no subset relation, and the whole-text token overlap is 0.33 — just under the 0.35 near-duplicate gate. The comment in `resolveDuplicate` already names the gap: *"a pair left coexisting is invisible afterwards."*

## Why the detector is lexical, not semantic

Similarity statistics were measured for exactly this job and lost: the motivating reversal scores **0.8593** under the default profile while a hand-labelled negative pair sits at **0.8575**, and five statistics (margin, ratio, margin-over-sd, z, exponential tail) all discriminate worse than the raw cosine they tried to replace. No gate over any of them reaches this case at a shippable fire rate. What distinguishes a reversal is not how *close* the two texts are — it is that one of them *says* the other is over. That is a lexical fact, detectable deterministically, and it works in the fresh two-item store where the CSLS guard abstains by construction (`MIN_POOL_FOR_Z`).

## What fires, and the measured rates

A sentence that contains a reversal cue (`no longer`, `abandoned`, `superseded`, …) **and** names another active item's *distinctive* title tokens — at least two, covering at least half of them. Distinctive is judged against this store's own title-token frequencies (`distinctiveTitleCap`), the same corpus-relative reasoning that moved the guard from a fitted constant to a percentile.

Measured on a real 831-item store (all numbers replayable, script shapes in the test file):

| variant | flagged pairs |
| --- | --- |
| any 2 shared title tokens near a cue | 13,678 |
| + coverage ≥ half the title | 21 |
| + distinctiveness vs the store's own titles | **6** (<1% of writes) |

Honesty about the 6: on that store they are narrative mentions, not live contradictions — the store is kept with strict `supersedes` discipline. That is why every surface reports this as a **candidate** and quotes the cue sentence back, so a false fire is dismissed on sight. The advisory never blocks, and an explicit `supersedes` is never second-guessed (its target is excluded before the scan runs).

## Surfaces

- **Write result** (`knowl_store` / `knowl_ingest_atoms`, single + batch): a `reversal` advisory beside `nearDuplicate`/`governingDecision`, phrased as a question, carrying the incoming write's own sentence and the exact `knowl_update` retire call. After this change, store 2 above comes back flagged and `knowl_update` resolves it in one step.
- **`knowl conflicts`** (CLI + MCP): now lists three kinds — `declared` (unchanged), `polarity` (the pairs the write path's own polarity guard deliberately clamps to coexist and could never show again), and `reversalCandidates`. The repro pair appears here even if the write-time note was ignored.
- Cost: the cue gate is a handful of substring checks, so ordinary writes pay nothing; a cue write reads active `(id, title)` only (new lean `listActiveKnowledgeTitles`). The conflicts pairwise title scan costs 21ms over 831 items. Polarity pairs found on that same store: 0 — quiet on a well-kept store.

## Verification

- `tests/store/contradiction-visibility.test.ts`: the motivating pair pinned end-to-end through `storeKnowledgeItemDeduped` (advisory fires, both items stay active), same-sentence requirement, distinctiveness and coverage negatives, supersedes suppression, and polarity-pair visibility in the scan.
- Full suite: 361 files, 3,505 passed, 5 pre-existing skips. Typecheck, lint, `docs:check` clean.
- MCP repro re-run against the built artifact: store 2 returns the flagged note; `knowl_conflicts` returns the pair.

### DIFF
--- a/docs/evals/reversal-detector-recall.md
+++ b/docs/evals/reversal-detector-recall.md
@@ -95,3 +95,20 @@ reversals, which by definition leave both items active and so cannot appear in
 The two measurements agree from opposite directions, which is why the conclusion holds: on the
 declared population recall is 4%, and on the active population — where an undeclared reversal
 *would* live — all 25 fires are narrative mentions and none is a real contradiction.
+
+## Outcome
+
+The detector stays on the **write path** and was removed from **`knowl conflicts`**.
+
+The same 4%/8% rate reads differently on the two surfaces, and that is the whole of the
+decision. At write time it is one note, attached to the writer's own sentence, quoted back, on
+2.4% of writes — a reader who did not mean a reversal dismisses it in seconds, and the sentence
+is right there to judge. In an inspection command it is a *list*: an agent reads a list as a
+work queue, pages it, and acts on it out of context. 45 candidates with no true positive among
+them is worse than the empty list it replaces, because the empty list is at least honest about
+what the store knows.
+
+So `scanContradictions` returns polarity pairs only — exact by construction, since they are the
+pairs the write path's own guard deliberately creates and no other surface could show. That was
+always the load-bearing half: `knowl conflicts` read `conflictKey`/`conflictExclusive` and
+nothing else, set on 3 of 937 active items.
--- a/docs/evals/reversal-detector-recall.md
+++ b/docs/evals/reversal-detector-recall.md
@@ -1,0 +1,97 @@
+# The reversal detector on real reversals: 4 of 101
+
+Measured 2026-08-24 against this repo's own store (1,165 items, 1,034 active) while reviewing
+PR #180. Replayable: `npx tsx scripts/probe-reversal-recall.ts <repo-root>`.
+
+## Why this measurement exists
+
+PR #180 shipped a reversal detector with a precision number and no recall number. The review
+measured precision here — 25 fires on active items, **all** narrative mentions, zero live
+contradictions — but that is a noise floor, not an evaluation. With no positive to score
+against, "does this detector find reversals" was unanswered, and two candidate precision gates
+each separated all 25 negatives from the **one synthetic positive** in the test file. Fitting a
+constant to a single point is the failure [`preset-floor-sweep.md`](preset-floor-sweep.md) and
+[`query-coverage-probe.md`](query-coverage-probe.md) record for the relevance floor, so both
+gates were left out pending real positives.
+
+The positives were already in the store. Every `superseded_by_id` link is a reversal that
+genuinely happened — someone decided B replaces A. The same population cleared the polarity
+guard as 0-of-101 in the PR #133 review.
+
+## Population
+
+| | count |
+| --- | ---: |
+| real supersessions (self-join on `superseded_by_id`) | 114 |
+| …titles already in a subset relation (`sameSubjectTitle` true — the write path catches these) | 13 |
+| **…titles unrelated — the detector's actual job** | **101** |
+
+## The result
+
+| | of 101 | |
+| --- | ---: | --- |
+| superseder's content contains any reversal cue | 30 | 29.7% |
+| **detector fires — a cue sentence names the predecessor** | **4** | **4.0%** |
+| | | (13.3% of the cue-bearing subset) |
+
+Against 45 false fires among active items at the shipped gate. Precision ≈ 8%.
+
+## No gate setting rescues it
+
+Swept over the 101 real positives and every candidate pair among active items:
+
+| gate | real reversals caught | false fires |
+| --- | ---: | ---: |
+| **shipped: ≥2 shared AND ≥50% of title** | 4 (4.0%) | 45 |
+| ≥2 AND ≥40% | 4 (4.0%) | 78 |
+| ≥2 AND ≥33% | 5 (5.0%) | 122 |
+| ≥3 shared, no ratio | 6 (5.9%) | 47 |
+| ≥4 shared, no ratio | 3 (3.0%) | 23 |
+| ≥3 OR ≥50% | 6 (5.9%) | 69 |
+| ≥`ceil(sqrt(title))` shared | 5 (5.0%) | 61 |
+
+Nothing exceeds 6% recall. **This settles the two gates the review declined to add**: a
+cue-sentence cap of ≤10 tokens keeps 1 of the 4 real reversals, and symmetric coverage ≥0.5
+keeps **0 of 4**. Both would have bought their clean precision by deleting the feature.
+
+## Per-cue contribution
+
+| cue | real caught | false fires |
+| --- | ---: | ---: |
+| `supersedes` | 3 | 8 |
+| `no longer` | 1 | 27 |
+| `superseded` | 0 | 6 |
+| `replaced by` | 0 | 4 |
+| `abandoned`, `deprecated`, `reversed`, `obsolete`, `overturned`, `rescinded`, `retracted` | 0 | 0 |
+
+**Seven of eleven cues have never fired at all**, in either direction, on a 1,165-item store.
+`no longer` is 27:1 against — it is the register release notes are written in ("X no longer
+does Y"), which is a *report* of a change, not an assertion that a stored item is over.
+`supersedes` carries essentially all the signal, and carries it because a writer using that word
+in prose is usually also passing the `supersedes` field — the case the advisory already excludes.
+
+## Why the misses miss
+
+The gate needs one sentence to name **half** of the predecessor title's distinctive tokens.
+Knowl titles are sentences, not labels — the missed pairs carry 8, 11, 14, 18 distinctive
+tokens. Half of 11 in one sentence is close to unreachable. One miss lands a single token short:
+
+> `The relevance floor separates queries, not candidates` superseding
+> `An absolute relevance floor DOES separate on-topic from off-topic` —
+> 8 distinctive tokens, 3 shared, needs 4.
+
+Lowering the ratio to reach it costs 122 false fires for one more catch.
+
+## What this does and does not prove
+
+**Does:** the mechanism is sound on its designed case (the Postgres/SQLite pair is caught) and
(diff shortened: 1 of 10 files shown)


dat999zx:
Reviewed and verified against this repo's real store (1,165 items, 1,033 active). The design holds up — lexical over semantic is the right call for this, the advisory never blocks, and the tests are real regression pins rather than restatements. I pushed three fixes to the branch; details below so you can push back on any of them.

## The numbers don't replicate on this store

The doc comments cite the 831-item store this was tuned against. Re-measured here:

| | doc comment | this store (1,033 active) |
| --- | --- | --- |
| items carrying a cue | 148 | 129 (12.5%) |
| flagged pairs | 6 | **43** |
| write-path fire rate | "under 1% noise" | **25 items = 2.4%** |
| true positives | 0 of 6 | **0 of 43** |
| `knowl conflicts` latency | — | **1,796ms** |

All 43 are narrative mentions — release notes saying what a change stopped doing (`.env.example no longer count as sensitive paths`), items citing a `supersedes` id, changelog enumerations. So precision on a real positive is **unmeasured**: the only positive that exists anywhere is the synthetic one in the test file. I put that in the `detectReversal` doc comment rather than leaving the 6/<1% figures standing.

## What I fixed

**1. Perf — `1,796ms → 480ms`.** `scanContradictions` is O(n²) and both title predicates tokenize both sides internally, so the pair loop paid four tokenizations per pair: 1,467ms of the 1,796ms total, to find zero polarity pairs. Split `sameSubjectTokens` / `polarityTokensDiffer` out at the token-set boundary and precompute one set per item. The title-taking versions now call them, so there's still one definition of each rule.

**2. Noise — `43 rows → 25`.** The scan listed every title a cue sentence shared tokens with. One `3.0.2 recovery-and-trust plan` release-note sentence matched **11** unrelated titles by itself, because the gate covers a share of the *held title's* tokens and a long enumerating sentence contains many tokens by accident. A sentence asserts one reversal, so the scan now takes the best-covered title per cue sentence — the same rule `reversalAdvisoryForWrite` already applied on the write path. The two surfaces disagreeing was the actual bug. New test pins it.

**3. Claims.** Doc comments updated to the rates that replicate, and the O(n²) cost model is now stated in the module header.

## What I deliberately did NOT fix

I swept two precision gates and both work "perfectly," which is exactly why I left them out:

- **cue-sentence token cap:** true positive is 4 tokens, every one of the 43 negatives is ≥ 11. A cap of 10 → 0/43 false, true kept.
- **symmetric coverage** (shared / sentence tokens): true positive lands on exactly 0.500, highest negative 0.45. Requiring ≥ 0.5 → 0/43 false, with **zero margin**.

Each separates 43 negatives from a single synthetic positive — a constant fitted to one data point. That's the failure #182 documents for the relevance floor ("the fix is not a fourth number"), and this PR's own header makes the same argument for moving off a fitted constant. So the rate is stated in the docs rather than tuned away. If a genuine unresolved reversal ever shows up in a real store, that's the point to revisit it with more than one positive to fit.

Verified locally on the branch: build, typecheck, `docs:check`, **361 files / 3506 passed / 5 skipped**. CI green on the full matrix.

Holding this open rather than merging, since the per-sentence dedup changes what the feature reports and that's your call, not mine.

dat999zx:
Follow-up: I found the ground truth this was missing, and it changes the recommendation.

## The positives were already in the store

The first review measured precision (25 fires, all narrative, zero real) and stopped there, because there was no positive to score against. But every `superseded_by_id` link is a reversal that **genuinely happened** — someone decided B replaces A. That's the same population that cleared the polarity guard as 0-of-101 in the #133 review, and it gives 101 labelled positives whose titles are unrelated: exactly this detector's job.

Added as [`docs/evals/reversal-detector-recall.md`](docs/evals/reversal-detector-recall.md) with a replayable probe (`npx tsx scripts/probe-reversal-recall.ts <root>`). No behaviour change in this commit.

## Recall: 4 of 101

| | of 101 | |
| --- | ---: | --- |
| superseder's content contains any cue | 30 | 29.7% |
| **detector fires** | **4** | **4.0%** (13.3% of cue-bearing) |

Against 45 false fires at the shipped gate. **Precision ≈ 8%.**

## No gate setting rescues it

| gate | real caught | false fires |
| --- | ---: | ---: |
| **shipped: ≥2 AND ≥50%** | 4 (4.0%) | 45 |
| ≥2 AND ≥40% | 4 (4.0%) | 78 |
| ≥2 AND ≥33% | 5 (5.0%) | 122 |
| ≥3, no ratio | 6 (5.9%) | 47 |
| ≥4, no ratio | 3 (3.0%) | 23 |
| ≥3 OR ≥50% | 6 (5.9%) | 69 |
| ≥`ceil(sqrt(title))` | 5 (5.0%) | 61 |

Nothing exceeds 5.9%. This also settles the two gates I declined to add last round — good thing they stayed out: the ≤10-token cap keeps **1 of 4** real reversals, and symmetric coverage ≥0.5 keeps **0 of 4**. Both would have bought their clean precision by deleting the feature.

## Per cue

| cue | real caught | false fires |
| --- | ---: | ---: |
| `supersedes` | 3 | 8 |
| `no longer` | 1 | **27** |
| `superseded` | 0 | 6 |
| `replaced by` | 0 | 4 |
| `abandoned`, `deprecated`, `reversed`, `obsolete`, `overturned`, `rescinded`, `retracted` | **0** | **0** |

Seven of eleven cues have never fired in either direction on a 1,165-item store. `no longer` is 27:1 against — it's the register release notes are written in ("X no longer does Y"), a *report* of a change rather than an assertion that a stored item is over. And `supersedes` carries the signal largely because a writer using that word in prose usually also passed the `supersedes` field, which the advisory already excludes.

**Why the misses miss:** the gate needs one sentence to name half the predecessor title's distinctive tokens, and Knowl titles are sentences, not labels — the missed pairs carry 8, 11, 14, 18 distinctive tokens. One lands a single token short (`The relevance floor separates queries, not candidates` → `An absolute relevance floor DOES separate on-topic from off-topic`: 8 distinctive, 3 shared, needs 4). Lowering the ratio to reach it costs 122 false fires for one more catch.

**Honest limit on this measurement:** these are *declared* supersessions. The detector targets *undeclared* reversals, which by definition leave both items active and can't appear in `superseded_by_id`. It's a proxy. But the two measurements agree from opposite directions — 4% recall on the declared population, and on the active population where an undeclared reversal *would* live, all 25 fires are narrative.

## What I'd suggest

Not my call, and I've made no behaviour change beyond the two fixes in the previous commit. But laying out the options as I see them:

1. **Split the PR.** The polarity half and the `knowl conflicts` broadening are clearly good and independent — the command genuinely was reading only `conflictKey`/`conflictExclusive` on 3 of 937 items, and polarity pairs are exact by construction, so they're listed as contradictions rather than candidates. That half stands on its own.
2. **Drop `reversalCandidates` from `knowl conflicts`, keep the write-path advisory.** 45 false candidates in an inspection command an agent reads is worse than an empty list. At write time it's one note on 2.4% of writes with the sentence quoted, which is cheap to dismiss.
3. **Ship the reversal half as-is with the numbers documented.** Defensible — the eval is now in the tree, so nobody re-tunes it blind — but it's a feature at 8% precision and 4% recall.

I'd take 1 + 2. Happy to do the split if you want it; the fixes and the eval are already on the branch either way.

dat999zx:
Done — acting on the eval above.

**`detectReversal` stays on the write path. It is gone from `knowl conflicts`.**

The same 4%/8% rate reads differently on the two surfaces, and that turned out to be the whole decision:

- **At write time** it is one note, attached to the writer's own sentence, quoted back, on 2.4% of writes. A reader who did not mean a reversal dismisses it in seconds with the evidence right there.
- **In an inspection command** it is a *list*. An agent reads a list as a work queue, pages it, and acts on it out of context. 45 candidates with no true positive among them is worse than the empty list it replaced — the empty list was at least honest about what the store knows.

So `scanContradictions` returns polarity pairs only. Those are exact by construction: they are the pairs the write path's own guard deliberately creates, and this was the only surface that could ever show them again. That was always the load-bearing half of the PR — `knowl conflicts` read `conflictKey`/`conflictExclusive` and nothing else, set on **3 of 937** active items.

**Removed:** the CLI output field, the MCP block and its truncation line, the `knowl_conflicts` description clause (it now points a caller at the write result for reversals), the two docs rows, `ReversalCandidate`, and the per-sentence dedup that went with it.

**Kept:** `detectReversal`, `reversalCueSentences`, `titleTokenFrequency`, `distinctiveTitleCap`, `reversalAdvisoryForWrite` — the advisory is unchanged, including the batch path and the exclude set.

**Side effect:** the scan is **1,796ms → 79.5ms** on a real 1,033-item store, 22×. What remains is the precomputed polarity pass.

**Tests:** the conflicts-listing assertion became the decision pin (`scanContradictions` returns `polarity` and nothing else), and the per-sentence dedup test moved to the write advisory, which still has that behaviour. 12 tests in the file, all passing.

Verified: build, typecheck, `docs:check`, **361 files / 3506 passed / 5 skipped**.

Branch is now four commits — your feature, the perf/dedup fixes, the eval, and this split. The reversal work is all still here and still running where the measurement supports it; if you disagree with where I drew the line, the eval and probe are in the tree to argue with.

Issue #146, opened 2026-08-20
Relevance floor under-triggers, and `score` cannot be used as the fallback signal it points at


williamttruong:
Found while building `@knowl/opencode`, which injects retrieved memory into the system prompt on every turn. That makes "does this store actually hold the answer" load-bearing: a wrong yes spends ~1,200 tokens of the user's budget on an unrelated atom, unasked.

The `abstained` mechanism is exactly the right design for this and I'm now keyed on it. Two things make it less usable than intended.

## Repro

A store with **exactly two atoms**, both plain English, one about a billing service and one about a refund queue. Nothing else. Then five identical runs per query:

```
abstained[AAAAA]  notice[NNNNN]  topScore=0.98  recipe for sourdough bread starter
abstained[-----]  notice[-----]  topScore=0.98  kubernetes ingress nginx tls renewal
abstained[-----]  notice[-----]  topScore=0.98  zzzzz qqqqq xxxxx vvvvv
```

`A` = at least one row carried `abstained: true`, `N` = the stderr floor notice appeared.

## 1. The floor under-triggers

`recipe for sourdough bread starter` abstains 5/5. `kubernetes ingress nginx tls renewal` and `zzzzz qqqqq xxxxx vvvvv` abstain **0/5** — against a store that contains nothing but billing and refunds. The last one is not even words.

Deterministic, not flaky: 5/5 and 0/5 both reproduce exactly. I checked this specifically because a single non-reproducing run had made it look nondeterministic — that was my own stderr capture, not the engine.

## 2. `score` cannot be the fallback

The notice says *"Read `score` and judge"*, and the MCP tool description says a low top score means the best available match is weak. In this store the top score is **0.98 for all three queries above and for a genuinely on-subject query** (`billing service port health check`). The second row scores `0` in every case.

So on a small store the score appears to be rank-derived rather than query-derived, and carries no information a caller can act on. A consumer told to read it has nothing to read. I tried gating on it before finding `abstained` — a floor anywhere below 0.98 passes gibberish, and anywhere above it rejects perfect matches.

I have not checked where this crosses over as the store grows; on the ~1,600-atom knowl-cloud store scores do vary (0.47–0.99), so this may be specific to stores small enough that the vector arm contributes nothing.

## Why it matters beyond one plugin

[...]

Issue #169, opened 2026-08-24
The relevance floor cannot be tuned on granite-small-en-r2: on/off-topic cosine distributions already overlap


dat999zx:
Split out of #146, as agreed in that thread. #146's other half — publishing `cosine` alongside `score` — shipped; this is what remains, and it is a model-selection question rather than a threshold one.

## The measurement

Reproduced at 5.7.0 (`ddf2c90`) against a fresh two-atom store (one atom about a billing service, one about a refund queue), `granite-small-en-r2`, both embedded, five runs per query. Deterministic. Raw cosines through `rankKnowledge`; the floor for this model is **0.76**.

| query | best cosine | vs floor | verdict |
| --- | --- | --- | --- |
| `recipe for sourdough bread starter` | 0.6640 | under | abstains ✅ |
| `kubernetes ingress nginx tls renewal` | **0.7928** | **over** | **answers ❌** |
| `zzzzz qqqqq xxxxx vvvvv` | 0.7478 | under | abstains ✅ |
| `billing service port health check` | 0.9296 | over | answers ✅ |

A query with nothing in common with the store scores 0.7928 — above the floor, and only 0.14 below a genuine match.

## Why it is not a threshold problem

`docs/evals/per-model-floor.md` already recorded, at calibration time, **on-topic min 0.7637 against off-topic max 0.7644**. The distributions had already crossed, by 0.0007, when 0.76 was chosen. The floor was placed inside the overlap because there was no gap to place it in.

No scalar threshold separates on-topic from off-topic on this model. Moving 0.76 anywhere trades a false answer for a false abstention one-for-one. And this is the **default** preset.

For contrast, from the same sweep:

| preset | on-topic min | off-topic max | gap |
| --- | --- | --- | --- |
| `arctic-embed-m-v2` | 0.1638 | 0.2275 | inverted, but wide |
| `granite-small-en-r2` *(default)* | 0.7637 | 0.7644 | **0.0007** |
| `granite-97m-multilingual` | 0.7443 | 0.7552 | 0.0109 |
| `bge-small-en` | 0.5399 | 0.5754 | 0.0355 |
| `minilm-l6-en` | 0.2003 | 0.2392 | 0.0389 |

granite-small compresses everything into a narrow high band. Every other preset has more room than the default does.

## Why it matters

[...]

PR #185, opened 2026-08-24
fix(transcripts): the cosine was computed and then thrown away by the fusion

### LINKED ISSUE #183: knowl_transcript_search has no relevance floor: nonsense queries return hits
## The defect

`knowl_transcript_search` returns hits for input with no semantic content at all. Reproduced against a live `knowl serve` over MCP stdio:

```
query: "zzzzz qqqqq xxxxx vvvvv wwwww"
→ returns transcript hits
```

`knowl_query` already solves this — it carries `belowRelevanceFloor` and an absolute `cosine` on a fixed scale, precisely so a caller can tell "the best available match is genuinely weak" from "this is the answer". `knowl_transcript_search` has neither, so nearest-neighbour noise is indistinguishable from a real recall.

## Why it matters more than tidiness

It spends agent context at exactly the wrong moment. Transcript search is reached *after* a memory miss — when the agent is already lost — and handing it plausible-looking irrelevant history there is worse than an honest empty result.

It is also the single blocker to a feature that otherwise looks attractive: **auto-chaining the transcript leg onto a `knowl_query` miss.** We backtested that idea before proposing it, and it fails today *because* of this defect:

- **40 distinct historical query misses** replayed through the real handler (`knowl serve` over stdio, not a reimplementation), with a hindsight filter so a hit only counts if its transcript message predates the miss: **0/40 recovered.** All four pre-event hits were lexical collisions — e.g. "hash chain audit log transparency proof" matched a tutoring session about SAT error logs.
- **Control arm, 15 healthy queries:** 7 would have had transcript matter appended in an always-on variant. Every graded one was junk (subagent prompt preambles, `ls` output, task-notification blobs — long generic lines that match anything).

With a relevance floor, that experiment becomes worth re-running rather than dead.

## Suggested shape

Mirror the `knowl_query` design rather than inventing a second vocabulary:

- an absolute similarity on a stable scale, comparable across queries
- a `belowRelevanceFloor` flag when the top hit does not clear it
- returning **zero hits with an honest notice** is a better answer than a nearest-noise list

## Verification

The backtest harness re-runs as-is against a patched build to confirm both arms move in the right direction (misses recovered up, healthy-query noise down): `scripts/memory-eval/backtest-transcript-leg.mjs` in `knowl-cloud`.

Found while measuring memory retrieval end-to-end on 2026-08-24. Four related auto-capture/auto-retrieval proposals were tested and killed the same day; this is the only one that survived contact with the data.

dat999zx:
Closes #183.

`knowl_transcript_search` returned hits for `zzzzz qqqqq xxxxx vvvvv wwwww` and
nothing marked them, so nearest-neighbour noise was indistinguishable from a real
recall. The number needed to tell them apart already existed.

`semanticRank` computes a real cosine, then `fuseRankings` replaces every score
with a Reciprocal Rank Fusion total built from POSITIONS -- so the top hit scores
about the same whether it was a perfect match or the least bad of five bad ones.
The absolute number was computed and discarded at the boundary.

THIS IS THE THIRD INSTANCE OF ONE PATTERN, not a third unrelated bug. #146: the
semantic half folded into a rescaled `relevance`, fixed by publishing `cosine`.
#181: `queryCoverage` multiplied into a page-normalised `raw/best`, fixed by
publishing `coverage`. Now the transcript fusion. Every time: an absolute number
computed internally, folded into a page-relative one, and only the page-relative
one handed to the caller.

`TranscriptHit.cosine` now survives the fusion. `fuseRankings` merges it across
arms rather than reordering them -- the lexical arm is pushed first and wins the
key, so a message found by both halves would otherwise keep the lexical copy and
lose the cosine. `judgeRelevanceFloor` is one exported rule because federation
fuses a second time across repos and two copies would drift invisibly.

Judged on the BEST hit, not per hit: that is the only claim the floor was
measured to support, and per-hit would cut the tail of a genuinely good recall --
the mistake knowl_query's floor already made once and was measured making. Rows
with no cosine are SKIPPED, never scored zero; absence is unjudged, not adverse,
and reading it as a low score is how a half-indexed archive would call every
query off-subject.

THE TWO CALLERS ARE TREATED DIFFERENTLY, and that is the point.

  - The TYPED tool returns weak hits with the notice attached. The caller asked
    for whatever is closest and can read the bodies.
  - The `search.transcripts.fallback` chain gets `confidentOnly: true` and
    withholds an off-subject page as a verified negative. Nobody asked for that
    search, and it runs at the moment the agent is already lost.

The issue's own backtest is why the fallback filters rather than flags: across 40
replayed real misses it recovered 0, every pre-event hit a lexical collision, and
7 of 15 healthy queries would have had junk appended.

WORDING FOLLOWS #182, NOT THE ISSUE. #183 describes knowl_query's floor as
telling a caller "the best available match is genuinely weak" from "this is the
answer". #182 corrected exactly that claim hours earlier -- the floor detects
off-subject, not unanswerable -- so the notice here says the narrow thing and
points the reader at the bodies. The issue's evidence survives that correction
untouched: gibberish and off-domain collisions are precisely the class the floor
does catch.

Verified: build, 362 files / 3514 passed / 5 skipped, typecheck, docs:check.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -520,6 +520,22 @@ canonical tool table below:
   `transcript://<repo>/<session>#L<line>` locators.
 - **`knowl_transcript_read`** — open one locator with the surrounding turns.
 
+Transcript search carries the same relevance verdict `knowl_query` does, and it means the same
+narrow thing: **off-subject, not unanswerable** (see [above](#what-abstained-means-and-what-it-does-not)).
+When a semantic half ran and the best hit falls below the model's floor, the results arrive under a
+`NO CONFIDENT MATCH` notice rather than silently — before this, a query of pure gibberish returned
+plausible-looking hits with nothing marking them.
+
+The two callers are treated differently on purpose:
+
+- **You typed `knowl_transcript_search`** — you asked for whatever is closest, so weak hits are
+  returned with the notice attached and you judge the bodies.
+- **The `search.transcripts.fallback` chain ran it for you** after a `knowl_query` miss — nobody
+  asked, so an off-subject page is withheld and reported as a verified negative. Appending
+  nearest-neighbour noise there spends context at the moment the agent is already lost, and it was
+  measured doing exactly that: across 40 replayed real misses it recovered **0**, while 7 of 15
+  healthy queries would have had junk appended.
+
 Disabling the feature **deletes** `.knowl/transcripts.db`. An index nothing will refresh is not
 something to leave on the disk of the person who just turned it off.
 
--- a/src/mcp/tools.ts
+++ b/src/mcp/tools.ts
@@ -1207,6 +1207,9 @@ export function registerTools(
               query,
               repos: Array.isArray(repos) ? repos.map(String) : undefined,
               limit: TRANSCRIPT_FALLBACK_LIMIT,
+              // Nobody asked for this search, so it has to clear the floor to be worth the
+              // context it spends. The typed tool still returns weak hits with a caveat.
+              confidentOnly: true,
             });
             const negative = fallback.startsWith(NO_TRANSCRIPT_MATCHES_PREFIX);
             const bounded = fallback.length > TRANSCRIPT_FALLBACK_CHARS
@@ -1215,7 +1218,7 @@ export function registerTools(
             blocks.push({
               type: 'text',
               text: negative
-                ? 'RECALL CHAIN — VERIFIED NEGATIVE: the knowledge store missed and the transcript archive holds no match for these words either (coverage below). If the user believes this happened, it predates the archive or used different words — try knowl_transcript_search with other words before concluding, and state the negative truthfully rather than guessing.\n'
+                ? 'RECALL CHAIN — VERIFIED NEGATIVE: the knowledge store missed and the transcript archive holds no confident match for these words either (coverage below). Anything that scored as off-subject was withheld rather than shown, because nothing asked for this search — run knowl_transcript_search yourself to see it. If the user believes this happened, it predates the archive or used different words — try other words before concluding, and state the negative truthfully rather than guessing.\n'
                   + bounded
                 : 'RECALL CHAIN: the knowledge store missed, so the transcript archive was searched automatically with the same words:\n'
                   + bounded,
--- a/src/transcripts/federate.ts
+++ b/src/transcripts/federate.ts
@@ -5,7 +5,7 @@ import { resolveStorage } from '../store/storage-roles.js';
 import type { ActiveWorkspace } from '../workspace/resolve.js';
 import { isTranscriptSharingEnabled } from './config.js';
 import { openTranscriptDb, TranscriptIndexMissingError } from './database.js';
-import { fuseRankings, searchTranscripts, type TranscriptHit } from './search.js';
+import { fuseRankings, judgeRelevanceFloor, searchTranscripts, type TranscriptHit } from './search.js';
 
 export type FederatedTranscriptHit = TranscriptHit & { repo: string };
 export type TranscriptSkipReason = 'absent' | 'not-shared' | 'unreadable';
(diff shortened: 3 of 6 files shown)


williamttruong:
The design here is right and I'd keep all of it — carrying `cosine` through the fusion, judging the best hit rather than each one, skipping unembedded rows instead of scoring them zero, and especially the split between the typed tool and the `confidentOnly` fallback. My backtest is why that split is correct and I'm glad it landed that way.

**The constant is the problem, and it fails in opposite directions depending on merge order with #186.** I measured both states on a live 9,100-message archive.

## Today, on `main`: the floor fires on everything

Transcript cosines on this machine span **0.47–0.59** against `relevanceFloor` 0.76. Not some of them — all of them. I built essentially this same judgement before finding your PR and ran it against the real archive:

```
ABSTAIN  hits=4  "zzzzz qqqqq xxxxx vvvvv wwwww"      ← correct
ABSTAIN  hits=3  "transcript search relevance floor"  ← on-subject, in this archive
ABSTAIN  hits=5  "capture nudge shouldNudgeForSilence"← on-subject, in this archive
```

100% of queries judged off-subject. With `confidentOnly: true` that makes the fallback chain return a verified negative for **every** query it ever runs, and the typed tool attach `NO CONFIDENT MATCH` to every search.

## After #186: the floor fires on nothing

[#186](https://github.com/dat999zx/knowl/pull/186) fixes a quantization clip — `granite-small-en-r2` carries a rogue component near 0.70 against a 0.3062 threshold, so one component of 384 clipped and cost every stored vector 25% of its norm. That was suppressing the whole cosine scale. Rebuilt correctly, the same archive reads:

| | before #186 | after #186 |
| --- | --- | --- |
| transcript top-1 cosine range | 0.47 – 0.59 | **0.82 – 0.83** |
| `zzzzz qqqqq xxxxx vvvvv wwwww` | below 0.76 | **0.8324 — above** |

So after #186 lands, `belowRelevanceFloor` is `false` for everything including pure gibberish, and #183 is silently unfixed. **Whichever order these merge, the floor never discriminates.**

## Because the distributions overlap, on clean data

That is not a "pick a better number" problem. Re-running your #169 sweep against the correctly-quantized rebuild, junk built the way #169 builds it (technical register, absent subject, 15/15/8):

| statistic | real min | technical max | gap | on-topic buried |
| --- | --- | --- | --- | --- |
| `top1` | 0.8230 | 0.8324 | −0.0094 | 3/15 (20%) |
| `top1 − top2` | 0.0007 | 0.0262 | −0.0255 | 13/15 (87%) |
| `top1 − mean(top10)` | 0.0083 | 0.0274 | −0.0191 | 8/15 (53%) |
| `top1 − median` | 0.0836 | 0.1113 | −0.0277 | 2/15 (13%) |
| `z` | 3.7946 | 4.6240 | −0.8294 | 5/15 (33%) |
| CSLS | 0.0301 | 0.0670 | −0.0368 | 8/15 (53%) |

Every gap negative. This is #169's conclusion confirmed on clean data, which is a stronger version of it than either of us had — including that CSLS, the lead I raised there, fails too.

## What I'd suggest, and it's small

Nothing here needs redesigning. The mechanism is sound and `judgeRelevanceFloor` already returns `undefined` for "unjudged", which is exactly the right shape for this.

**A transcript corpus has no measured floor, so pass `null`** — the same withheld claim `relevanceFloorFor` makes for an unmeasured model, one corpus along instead of one model along. `MODEL_RELEVANCE_FLOORS` was calibrated on curated atoms; transcripts are a different distribution and were never in that calibration. Then the plumbing ships, the notice stays wired, and the constant arrives when someone measures one — at which point the tests you've already written are exactly the tests it needs.

If you'd rather ship a number now, the merge-order interaction with #186 is the thing to pin down first, because the two PRs move the scale by ~0.30 and 0.76 sits on opposite sides of it before and after.

Reproducible: `scripts/memory-eval/probe-transcript-floor.mts` in knowl-cloud, run from a knowl checkout against any project root.

Sorry for the noise on #183 — I filed it proposing a fix you'd already disproved in #169, then found the quantization bug while measuring why my version of it didn't work. Your PR is the better answer to the issue.

PR #222, opened 2026-09-01
fix(context): a correction elsewhere was promoting the atom it made doubtful

williamttruong:
`getRecentContext` orders the session card's three knowledge slots by `updated_at`. That column moves on supersession, archival, visibility promotion and a freshness flip — `unrestated.ts` already documents this, and measured **72% of items** on a 950-item store carrying an `updated_at` newer than their `valid_from`.

## The consequence was specific, and backwards

`blastRadius` marks a sibling `needs_review` through `updateKnowledgeItem`, which stamps `updated_at` unconditionally (`repository.ts:443`). So correcting one atom **promoted unrelated atoms onto the next session's card — for having just been flagged as doubtful** — and evicted whatever that session had actually learned.

A card has three knowledge slots. Housekeeping could take all three.

The same applies to `gc.ts` archiving, to `knowledge-writer.ts` retiring a predecessor, and to a visibility promotion.

## The fix

Order by the open assertion's `valid_from`. A new assertion generation is written only when title, content, reasoning or confidence change, so it moves on **restatement and on nothing else** — the property `unrestated.ts` relies on, using the same join, so the two surfaces cannot disagree about what "recent" means.

**LEFT joined, where that module inner-joins.** It reports on a population and may exclude a row it cannot date; this is a card with three slots, and an item with no open assertion — an import, a store predating the table — must still be reachable. Such a row coalesces to `updated_at`, which is exactly the previous behaviour for exactly the rows that have nothing better.

## Two tests, and the second matters as much as the first

- A `needs_review` flip must **not** move an atom onto the card.
- A genuine content change must **still** move it there.

Without the second, this would have traded a wrong signal for no signal.

## Verification

| Gate | Result |
| --- | --- |
| `npx tsc --noEmit` | clean |
| `npm run lint` | clean, 0 errors |
| `npm run build` | clean |
| `npx vitest run` | **374 files / 3578 passed / 5 skipped / 0 failed** |

**Falsification:** restoring the `updated_at` ordering fails exactly the two new tests. The existing recency test (`should return recent active knowledge and commits…`) passes either way, because its fixture's creation order and its update order agree — worth knowing, since that is why the defect survived having a test over this function.

Note for whoever runs the suite: `npm run build` first, or the integration suites spawn a stale `dist/index.js`.

### DIFF
--- a/src/store/recent-context.ts
+++ b/src/store/recent-context.ts
@@ -1,4 +1,4 @@
-import { and, desc, eq, notLike } from 'drizzle-orm';
+import { and, desc, eq, isNull, notLike, sql } from 'drizzle-orm';
 import { KnowledgeCommit, KnowledgeItem } from '../core/types.js';
 import { DatabaseError } from '../core/errors.js';
 import { getDb } from './database.js';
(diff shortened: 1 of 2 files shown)


williamttruong:
CI caught a real defect the local suite could not, and it is worth naming because it was not in the change I opened this for.

`does not promote an item onto the card` passed on Windows and ubuntu and **failed on the macOS runner**, first slot right and second slot wrong. Cause: ISO timestamps are millisecond-granular, so three atoms written by one turn routinely share a millisecond on a fast machine — `valid_from` alone is a **partial** order, and rows tied inside it came back in whatever order the engine picked.

This is the same class `tier.ts` already documents for `tier_since`:

> ISO timestamps are millisecond-granular — so a confirmation recorded in the same millisecond as the boundary was dropped … on CI's ubuntu runner three tests in this area failed at random depending on which side of a tick they landed.

**`updated_at` had the identical property**, so the instability is not introduced here — the new test is just the first thing sensitive enough to surface it. Two fixes pushed:

1. **The ordering is now total**: `valid_from`, then `created_at`, then `id`. A card that can differ between two identical calls is instability that reads as a bug in whatever consumed it.
2. **The fixture pins the clock**, via a `setAssertionValidFrom` helper beside the `setKnowledgeItemUpdatedAt` that already pins the other one.

The restatement test is **rewritten rather than pinned**. It asserted a card *position* after a restatement, and the new assertion generation is stamped `now` — which cannot be ordered against a pinned future instant without the fixture pinning the very value under test, and cannot be ordered against a pinned past one either, since this file shares one store across all its tests. It now asserts the mechanism the ordering rests on, in both directions at once: a content change moves `valid_from`, a `needs_review` flip does not.

Re-verified: 374 files / 3578 passed / 0 failed, lint and typecheck clean, and the store suite run three times over for the tie.

williamttruong:
The macOS failure is fixed and that runner is green. The remaining red — **ubuntu node 24** — is a **pre-existing flake unrelated to this PR**, and I chased it rather than re-running until it passed.

`tests/store/conflicts.test.ts > repairs keys already stored raw…` creates two rows back-to-back and asserts the second survives the repair. When both land in the same millisecond, `integrity.ts` falls through to its documented `a.id.localeCompare(b.id)` tiebreak and the survivor is whichever random id sorts first. Driving that fixture with the timestamps forced equal, over 12 fresh stores: **passes 3, fails 9.** The CI message was `expected '20b4015dc5304e1a' to be 'f73b82c66642461b'` — `2…` before `f…`, the tiebreak choosing exactly as predicted.

Nothing in this PR touches `integrity.ts`, `conflicts.ts`, or that test. Fix sent separately as #226 so the two stay reviewable apart; it merges cleanly with this branch.

For what it is worth, that is the same millisecond-granularity class as the macOS failure here and as the `tier_since` note in `tier.ts` — three places now. If you want a fourth thing to look at some day, a shared test helper that pins these clocks would probably retire the category.

Issue #243, opened 2026-09-03
Hermes host: shell hooks are never registered by Hermes Desktop (serve backend) — 0/6 on Desktop, terminal-only until upstream #69825


williamttruong:
## Summary

`knowl init hermes` (5.19.0, #240) wires Knowl into Hermes via seven `config.yaml` shell hooks. Hermes **Desktop** never registers those hooks, so the host delivers 0/6 capabilities for every Desktop user and only works from `hermes chat` in a terminal. `hermes hooks doctor|test` report healthy because they read config, not the live registry.

Verified against the installed Hermes v0.21.0 source (`hermes-agent` @ 6327930, 2026-09-03). Upstream: NousResearch/hermes-agent #41457 (2026-06-07), #69825, #93998 (re-confirmed on 0.20.5, 2026-08-24). Open, 14+ competing PRs, none merged.

## Evidence (file:line in hermes-agent 0.21.0)

- Desktop spawns `python -m hermes_cli.main serve --host 127.0.0.1 --port 0` (`apps/desktop/electron/backend-command.test.ts:16`).
- `hermes_cli/main.py:12998` `_try_fast_serve_launch()` calls `cmd_dashboard(args)` directly. The CLI's shell-hook registration lives in `_prepare_agent_startup()` (`:12921-12924`), reached only from chat paths.
- `cmd_dashboard` -> `hermes_cli/web_server.start_server` -> `tui_gateway/`. `grep -rn register_from_config tui_gateway/ hermes_cli/web_server.py` is empty apart from the dashboard's hooks list/approve/revoke UI. The agent is built in-process (`tui_gateway/server.py:9321`) or in a compute-host child whose `main()` only calls `run_host()` (`tui_gateway/compute_host.py:942-946`).
- All `register_from_config` call sites: `cli.py:1060`, `hermes_cli/main.py:12924`, `gateway/run.py:14166`/`:17276` (messaging gateway, not Desktop), `agent/shell_hooks.py:373`.
- `hermes_cli/hooks.py:56, :273, :353, :385` use `iter_configured_hooks(load_config())` + the allowlist. The registry is never consulted.
- Python plugins DO load in Desktop: `tui_gateway/server.py:6281-6283` `discover_plugins()`.

## Suggested handling

1. Docs: change the hosts matrix row from "✅ edit turns" to "terminal only until upstream #69825 lands"; add a warning to the `knowl init hermes` output.
[...]

PR #244, opened 2026-09-03
feat(hosts): Hermes plugin so Knowl reaches Hermes Desktop

williamttruong:
Closes the Desktop half of #243.

## Why

Hermes Desktop's `serve` backend never registers `config.yaml` shell hooks (NousResearch/hermes-agent #69825, still open at v0.21.0). Confirmed live: the same marker hook fires on `hermes chat` and fires nothing on a backend launched with Desktop's exact command, while `hermes hooks doctor` reports healthy because it reads config, not the live registry. So `knowl init hermes` is terminal-only today. Python plugins do load on that path.

## What

`integrations/hermes/` (beside `integrations/cline/`): a Hermes Python plugin, `knowl/plugin.yaml` + `knowl/__init__.py`, ~350 lines, that shells to `knowl agent-hook hermes <event> --json` with the shell-hook payload shape, so `src/session/hosts/hermes.ts` keeps doing all the memory work. It adds what shell hooks cannot: the correct per-session cwd on Desktop, a frozen system-prompt rules section, a same-turn impact card on file writes via `transform_tool_result`, and context bounded under Hermes' 10k spill. It does not forward `on_session_start` (see the bootstrap-loss note on #243; the shell-hook profile has the same loss and dropping that event from `HERMES_HOOK_EVENTS` is the matching fix there).

Also: `install.py` (directory install), `pyproject.toml` (pip install via the `hermes_agent.plugins` entry point, same `knowl` key), `verify/drive.mjs` (drives a `serve` backend over JSON-RPC for an end-to-end check), a Hermes paragraph in `docs/hosts.md`, and a changelog entry. `integrations/` already ships in the npm package with `__pycache__` excluded.

## Verified (Hermes v0.21.0 @6327930, Windows 11, knowl 5.19.0)

| Run | Result |
|---|---|
| Memory-only question | Haiku 4.5: "Port 4321, internal only—never expose publicly. From Knowl project memory (recent constraint, updated Sept 3)." No file read. |
| Write contradicting a stored constraint | Refused and asked, citing the constraint from the recall card |
| Constraint-respecting write | `write_file` result gained `[Knowl] 1 stored item(s) depend on README.md ...`; model quoted it |
| Entry-point install | `uv pip install` of the wheel into Hermes' venv → `hermes plugins list`: `knowl \| enabled \| entrypoint` |
| Plugin warnings/tracebacks, five runs | none |

## Gates

No TypeScript touched. `git diff --check` clean. `npm test` not run for this change; the only files under `src/`/`tests/` are untouched.

## Left for you

- `knowl init hermes` installing this plugin (copy from the npm package's `integrations/hermes/`) instead of, or beside, the shell hooks
- the bootstrap-loss fix in `hosts/hermes.ts` (#243)
- publishing `knowl-hermes` to PyPI is William's


dat999zx:
Reviewed and built on in #245 (this branch's two commits are included there, authorship intact) — happy to close this in favour of that, or land this first and rebase.

**The finding holds up.** Verified independently against the installed v0.21.0: `web_server.py` touches `shell_hooks` only for the dashboard's approve/revoke list, `tui_gateway/` never mentions it, and `discover_plugins()` runs from `agent/agent_init.py` — which every path builds an agent through. So the plugin isn't a Desktop side-channel, it's the channel, and #245 makes `knowl init hermes` install it and **remove the shell hooks 5.19.0 wrote** (both registered = every event twice in a terminal session).

**Bootstrap loss, confirmed and fixed in the engine.** Your note was right and it applied to the shell-hook path too. Measured: fresh session, `pre_llm_call` first → 3,030-char card; preceded by `on_session_start` → empty on both. The event is no longer registered, so the first turn binds and carries the card.

**Three bugs found in the plugin:**

1. `query()` sliced from the first `[`. `knowl query` returns a bare array in a lone repo but a **repo-keyed object** in a linked workspace, so `json.loads` threw "Extra data" and returned `[]` — the impact card was silently dead in every repo with neighbours, including knowl itself. That's why it verified clean for you: a lone repo takes the other branch.
2. `pre_tool_call` used the same 30s Hermes bounds it at, and it's the one hook in `_HOOK_TIMEOUT_FAIL_CLOSED_HOOKS` — so a slow gate was a race whose loser is the user's write blocked with a reason nobody wrote. It now answers inside Hermes' own configured bound, with a margin.
3. Minor: the dedupe set grew unbounded in a long-lived Desktop backend, and `lstrip("./")` strips a character *set*, collapsing `.github/x` onto `github/x`.

Added the 16 unittest cases the plugin shipped without — the workspace-shape bug is exactly the kind a test catches.

Everything else in the plugin checked out against source: kwargs dispatch filtered to declared params (so the missing `conversation_history` is harmless), `register_system_prompt_section` and `transform_tool_result` signatures match, and `resolve_context_cwd` is the right per-session cwd. Nice work — the Desktop diagnosis is the part that would have taken anyone else a long time to find.

dat999zx:
Superseded by #245, which contains these commits with authorship intact and adds the review fixes on top.

PR #255, opened 2026-09-04
feat(code): Python and Go in the symbol index

### LINKED ISSUE #254: Runtime deps: tree-sitter-python and tree-sitter-go, so the code index reads Python and Go
## What

Add `tree-sitter-python` and `tree-sitter-go` as runtime dependencies, so the code index behind `symbol://` locators, the per-symbol read set and certain-tier impact detection cover `.py` and `.go` files as well as `.ts`/`.tsx`/`.js`/`.jsx`.

CONTRIBUTING asks for a discussion before any new runtime dependency, so this is that discussion. The implementation is ready and I'll open the PR against this issue.

## Why

Today a `.py` or `.go` file is skipped at both index entry points (`walkCodeFiles` and `indexFile`) without a word. A session that reads one records a single `file://` row, so a comment-only edit reports it as moved — file granularity on a tier the plan holds to ≥95% precision. Mixed repos (TS front-end, Go/Python services) get the feature for half their code.

`docs/reference.md` is honest about the limit ("supports `.ts`, `.tsx`, `.js`, and `.jsx`"), and `docs/change-impact-plan.md` §7.3 already scopes the approach to shallow tree-sitter — this stays inside that decision. Everything downstream of the extractor (schema, `CodeSymbolKind` as opaque text, `impact.ts`, evidence staleness, drift) is already language-agnostic; the language-shaped code is `src/code/languages.ts` and the walk in `symbol-index.ts`.

## The dependencies

| package | version | peer on `tree-sitter` | prebuilds |
|---|---|---|---|
| `tree-sitter-python` | `^0.23.4` | `^0.21.1` ✓ | darwin/linux/win32 × x64/arm64 |
| `tree-sitter-go` | `^0.23.4` | `^0.21.1` ✓ | darwin/linux/win32 × x64/arm64 |

- Same family and install mechanism (`node-gyp-build`) as the two grammars already in `dependencies`; nothing compiles at install on any platform CI runs.
- `tree-sitter-python@0.23.5+` moves its peer to `^0.22.1`, so `^0.23.4` cannot float onto an incompatible release under the current `tree-sitter@^0.21.1` runtime. `tree-sitter-go` is `^0.21.1` across all of 0.23.x.
- ~1.4 MB unpacked combined. Both need adding to `tsup.config.ts` `external`, as the existing grammars are.
- Not proposed as `optionalDependencies`: that pattern exists for `@huggingface/transformers` because it pulls a compiled ONNX runtime and downloads weights; these are two small prebuilt addons, and making them optional would force `languageForExtension` from a synchronous lookup into a lazy registry for no install-risk gain.

## Not in scope

No LSP / type-aware resolution — §11 of the change-impact plan already rules that out unless P-4 precision fails because of resolution depth.

Happy to adjust versions, make them optional, or drop one language if you'd rather take them separately.

dragospopa420:
Closes #254.

## What

The Tree-sitter index behind `symbol://` locators, the per-symbol read set and certain-tier impact detection covered `.ts`, `.tsx`, `.js` and `.jsx` only. A `.py` or `.go` file was skipped at both entry points without a word, so a session that read one recorded a single `file://` row and a comment-only edit reported it as moved.

Both languages now index:

- **Python** — `def`/`class` (decorated or not), class members as `Class.member`, module-level assignments, both import forms.
- **Go** — functions, methods as `Receiver.Method` (base type under generics, past a receiver comment), `type` specs and aliases under a new `type` kind, `const`/`var` specs of the declaration alone, every import path.

## How

`extractSymbols` moves out of `symbol-index.ts` into `src/code/extract.ts`. The JS/TS walk is unchanged — verified by indexing this repo's own `src/` and `tests/` on `main` and on this branch and diffing every `code_symbols` and `code_symbol_edges` row: no difference outside the files this PR touches, so no existing locator, signature or hash moves on upgrade. The Python and Go walks sit beside it, dispatched by extension from one `GRAMMARS` record in `languages.ts`.

A signature is the header before the body, taken from the grammar's `body` field rather than the first `{` — a `chan struct{}` parameter or a `[T interface{ M() }]` constraint would otherwise cut it. A struct or interface ends at its keyword, like a class at its brace, so a field comment does not move the hash and a parameter, decorator or receiver change does.

`__pycache__`, `.venv`, `venv` and `vendor` join `EXCLUDED_DIRECTORIES`. `indexFile` now forgets rows under a refused directory instead of leaving them frozen until someone runs `knowl index-code` by hand, so the two entry points agree about what is indexed; the CHANGELOG says what that means for a first-party directory by one of those names. `go.sum`, `poetry.lock`, `uv.lock` and `Pipfile.lock` join the churn paths drift ignores.

## Dependencies

`tree-sitter-python@^0.23.4` and `tree-sitter-go@^0.23.4`, per #254: same family as the two grammars already present, peer `^0.21.1` matches the installed runtime, prebuilds for every platform CI runs, 0.23.4 is the ceiling for Python under the 0.21 peer. Both added to `tsup.config.ts` `external`.

## Verification

- `npm test` — 402 files, 3817 passed, 0 failed
- `npm run lint`, `npm run typecheck`, `npm run build`, `npm run docs:check`, `check:versions`, `git diff --check`
- `tests/code/symbol-index-languages.test.ts` pins every shape above, including the body-edit-vs-signature-edit hash behaviour, the field-comment case, the function-literal-locals case and the forgotten-rows case.
- Smoke-tested against a 94 KB real-world Go file (14 ms) and a 143 KB Python file.

## Known recall gaps, on purpose

`a = b = 1` records `a` only; definitions under a module-level `if`/`try` are not indexed. Both match how the JS walk treats inner declarations, and a later fix only adds locators.


dragospopa420:
I have read the Knowl Contributor License Agreement and I agree to it.

PR #279, opened 2026-09-09
feat(workspace): dissent — contest a linked repo's item without editing it

williamttruong:
## What

A repo can record that a **linked repo's** item is wrong, and from that moment every query returning that item — in any repo in the workspace — carries the dispute.

```
knowl dissent record 5fdbe969633c418d --claim "The TTL is five minutes; measured on staging."
```

Agents get `knowl_dissent` in a linked repo (`record | list | reject | withdraw | reopen`), and a `disputed` block on every `knowl_query` result — search and id-fetch — that names the repo contesting it and the claim.

## Why

One repo owns an item and only that repo may change it. That rule is right — it is what stops a neighbour silently retiring knowledge its owner had good reason to hold. But on its own it leaves the repo that *noticed* the error with nowhere to put what it found: the write is refused and the finding is lost.

This is the resolution recorded in the blocked workspace-v2 decision, way out #2 — *"design an explicit proposal/accept authorization workflow — feature-sized, does not exist."* That decision blocked v2 for promising cross-owner editing and forbidding it in the same document, because the only shape considered was one repo reaching into another's rows. This is the other shape.

## The design in one line

**Neither repo ever writes the other's database.** The dissent lives in the store of the repo that raised it; the resolution lives in the store of the repo that owns the item; they are joined only when something is read. `assertOwnedItem` is untouched and `tests/mcp/foreign-item-refusal.test.ts` passes unchanged.

## The product decision worth arguing about

**A dissent marks the atom at the point of use, rather than sitting in a queue.**

The alternative — record it, notify the owner, change nothing until they act — was considered and rejected. It leaves the system knowingly serving a fact one of its own repos has flagged as wrong, at full confidence, until somebody happens to open a queue. It is also the git-PR model minus the part that makes git work: a PR blocks the merge, so the wrong code does not ship while it is open. A dissent blocks nothing.

Marking is bounded in a way superseding is not. The worst case is that a correct atom reads as *disputed* — it is still returned, still readable, and the owner can reject the dissent. A supersede makes the atom vanish from results. Those are not the same kind of harm.

It is also what the rest of the codebase already does. Supersession retires the predecessor so the wrong one stops being returned; the cross-repo overlap detector fires on every write rather than into a report; drift marks atoms stale; `pathsChanged` says "verify" on the row itself. The house answer to "when should the user learn something" is *at the point of use*.

## Ending a dispute

- **Accept** — supersede the item, as the owner always could. The dispute clears itself because a retired atom is not returned. **There is deliberately no accept verb.**
- **Reject** — the item stands and stops reading as disputed. The other repo keeps its own record of disagreeing; rejecting answers it rather than deleting it.
- **Rewrite** — a dissent is pinned to the revision it was raised against, so the owner rewriting the item stales it out. The objection was to what the item *said*.
- **Withdraw** — the raising repo takes it back.

`reject` has an inverse, `reopen`, and deliberately from the start. A rejection is a judgement made on partial information — the other repo is the one that saw the problem — and shipping it irreversible would repeat the shape of `workspace promote`, whose missing inverse has cost something every day since. It is safe precisely because the resolution row is local and additive.

## Notes for review

- **`KNOWL_MIGRATION_LEVEL` 16 → 17**, and it is load-bearing. `annotateDisputes` reads both tables on every workspace query, so a store left at 16 would skip `SCHEMA_STATEMENTS`, never create them, and fail *ordinary reads* — the failure level 6 records for `knowledge_forget_log`. `KNOWL_SCHEMA_VERSION` stays at 1. The schema pin for 17 is `7a690fc55d186b8106ea27c647685bd3`.
- **The overlay reads both repos' stores.** A third repo holds neither half, so reading only locally would leave every settled dispute looking open to everyone except the two repos involved. Pinned by a test that rejects from `b` and reads from `a`.
- **Staleness is judged against the owner's current row**, not the item passed in — callers hand the overlay compact search results whose content is truncated by design, and hashing that would make every atom read as rewritten.
- **No ranking change.** The overlay annotates and preserves input order exactly, pinned by a test. Demoting on dissent would let a neighbour push another repo's knowledge down the ranking — the blast radius the single-owner rule exists to prevent, arriving by a quieter route. If disputed atoms should rank lower, that is a separate change with its own measurement.
- **Cloud-owned items are refused.** A dissent against a team-replica row would sit on one machine forever, invisible to whoever could act on it, while reading as though something had been done. **This is what keeps the server-side impact of this PR at exactly zero.**
- **`dissents` and `dissent_resolutions` are `preserved` in snapshots**, beside tombstones and the forget log: restoring older knowledge is not a statement that anyone changed their mind, so `restored` would resurrect withdrawn dissents and un-reject settled disputes.
- **The id-fetch path resolves the workspace on a local hit too**, where #109 resolved it only on a miss. A local item can be disputed, and its owner fetching it whole is exactly who needs to be told. The namespace-layered lookup from #261/#264 is kept as-is underneath.
- **One tool more when linked into a local workspace** — README's count moves from 1 to 2 on that clause; the base 28 is unchanged.

## Server / knowl-cloud

**Zero changes, and by construction rather than by luck.** The team replica is a peer with no `dissents` table, which the overlay treats as empty; and cloud-owned targets are refused outright. If dissents should ever cross machines, that lands at the recorded sync-contract seam — dissent and resolution row types in `packages/contract`, a server table with RLS — and the rule at review time is: never land the server half without the client caller.

## Open, for you

Reader-role dissent rights if this ever syncs; rank demotion (only with measurement); whether dissents belong in `knowl export`.

## History

Built 2026-08-15 stacked on #109 and opened, by mistake, against the fork's own `feat/foreign-atom-read` branch rather than here — so it has been finished and invisible for three and a half weeks. Rebased onto `main` at 8e16c55 today; the two #109 commits it carried are dropped since #109 landed, and the seven dissent commits are unchanged in substance. Merge points, for the diff reader: the migration-level and schema-pin bumps, the `WORKSPACE` guidance bullet now sitting after the `repo:` one, the `pathsChanged` and `disputed` annotations coexisting in `tools.ts`, and `AGENTS.md` left untracked per #172.

## Verification

On the rebased branch at its head, after `npm ci` and `npm run build`:

| Gate | Result |
|---|---|
| `npm test` | **424 files passed, 4000 tests passed, 5 skipped, 0 failed** |
| `npm run typecheck` | clean |
| `npm run lint` | clean |
| `npm run docs:check` | "Generated documentation regions are current, and every tool and command is documented." |
| `tests/store/schema-pin.test.ts` | 8/8 — the level-17 pin is the value the test itself demanded, not a guess |
| `tests/mcp/foreign-item-refusal.test.ts` | unchanged and passing — the ownership rule is untouched |

New tests carried by the branch: `tests/workspace/dissents.test.ts` (529 lines: record, list, reject, reopen, withdraw, staleness against a rewrite, the both-stores overlay read from a third repo, order preservation, cloud-owned refusal) and `tests/mcp/dissent-tool.test.ts` (183 lines: the tool's gating on a workspace, each action's shape, and the `disputed` block on search and on id-fetch).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,47 @@ Notable changes to `@dat999zx/knowl`. Versions before 2.1.0 predate this file; s
 
 ## Unreleased
 
+### A repo can say a linked repo's item is wrong
+
+One repo owns an item and only that repo may change it. That rule is what stops a neighbour
+silently retiring knowledge its owner had good reason to hold — but on its own it left the repo
+that *noticed* an error with nowhere to put what it found. The write was refused, and the finding
+was lost.
+
+`knowl dissent record` is that place:
+
+```
+knowl dissent record 5fdbe969633c418d --claim "The TTL is five minutes; measured on staging."
+```
+
+Nothing in the other repo is touched. What changes is that every query returning that item — in
+any repo in the workspace — now carries the dispute beside it, so the next agent to rely on it
+reads the objection first instead of discovering it later. That is the whole point: a dissent that
+only sat in a queue would leave the system knowingly serving a fact one of its own repos has
+flagged as wrong.
+
+The owner sees it with `knowl dissent list --incoming` and ends it in one of two ways. **Accept**
+by superseding the item, exactly as they always could — the dispute clears itself, which is why
+there is no accept verb. **Reject** with `knowl dissent reject`, and the item stands and stops
+reading as disputed; the other repo keeps its own record of disagreeing, because rejecting answers
+it rather than deleting it. A rejection can be undone with `knowl dissent reopen`, since it is a
+judgement made on partial information and reconsidering has to stay possible.
+
+A dispute also ends quietly when the owner simply rewrites the item: a dissent is pinned to the
+revision it was raised against, because the objection was to what the item *said*.
+
+**Neither repo ever writes the other's database.** The dissent lives in the store of the repo that
+raised it and the resolution in the store of the repo that owns the item, joined only when
+something is read. The single-owner rule is not weakened anywhere — `knowl_update` on another
+repo's item is refused exactly as before, and now says what to do instead.
+
+Dissent is workspace-local. An item owned by a cloud workspace is refused rather than recorded,
+because that dissent could never reach its owner and would sit on one machine looking as though
+something had been done.
+
+Agents get `knowl_dissent` in a linked repo, and a `disputed` block on every query result that has
+one.
+
 **The OpenClaw bootstrap card was being thrown away, on every session.** The profile registered `session_start`, so the engine bound the session and spent the bootstrap card on an event whose return value OpenClaw discards — and the first real turn (`before_prompt_build`) then arrived on a session the engine had already seen, with nothing to say. Measured: a fresh session whose first event is `before_prompt_build` gets the orientation card; the same session preceded by `session_start` got an empty answer with zero prepend context. That event is no longer mapped to the engine lifecycle, so the first `before_prompt_build` binds the session and carries the card. The plugin continues to register `session_start` solely to warm the workspace handle cache in memory so the initial write gate avoids cold open latency.
 
 **A workspace is a path, not a prefix.** `OpenClawEngineManager` matched cached workspace handles by string prefix (`cwd.startsWith(root)`), so sibling directories that share a name prefix (such as `knowl` and `knowl-cloud`) collided on lookup: whichever warmed first captured subsequent requests for the other, reading and writing atoms to the wrong project's database. Workspace matching in `getHandle` and `releaseWorkspace` now enforces a path boundary via `path.relative`, ensuring a workspace matches only itself or directories beneath it.
--- a/KNOWL.md
+++ b/KNOWL.md
@@ -41,6 +41,7 @@ Casual conversation, a single memory lookup, and trivial non-resumable work do n
 - Whether a new write is shared is this repo's recorded default, not a fixed rule: joining a linked workspace sets it to workspace visibility, while `--default-visibility repo` keeps writes private. `knowl workspace set` with no flags prints the current value.
 - Knowledge already written privately stays private until someone runs `knowl workspace promote`. Only the owning repo can promote, update, or retire its own items.
 - When the work you are doing belongs to a LINKED repo, pass `repo: "<name>"` on the call instead of writing it here. The call then applies to that repo exactly as if it had been run there -- what you write is stamped as its own, and retiring its knowledge is allowed, because the repo is correcting itself. Without it a write lands here, attributed here, which is how a finding ends up in the wrong drawer.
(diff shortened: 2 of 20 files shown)


dat999zx (commented):
Reviewed in full at 163090d, with the suite and mutants run locally. Design holds: two single-writer rows joined at read, `assertOwnedItem` untouched, migration 17 additive and pinned, snapshot policy classified. Hot-path cost measured at ~0.4 ms/query with no dissents, ~4 ms with one; peers are pooled so the second store opens once per process.

**Two holes in the resolution side, fixed and pushed as 41468ab:**

1. `rejectDissent` (dissents.ts:313) checked only that the target existed locally. A repo holding a local copy stamped `origin_repo='b'` — exactly the shape `resolveTarget` accepts as a target — could reject a dissent against b's item. Now `assertOwnedItem` runs, and `workspace` threads through CLI and MCP.
2. `annotateDisputes` (:391-425) collected `dissent_resolutions` from every readable store into one set. A row in A's store silenced B's dispute for every reader. Resolutions now count only when they sit in the store of `target_repo`.

Both had tests that failed against the PR head as submitted.

3. The cloud-owned refusal was untested: `if (false)` on `heldByCloudReplica` left all 37 tests green. `tests/workspace/dissents-cloud.test.ts` now kills it.

**Not blocking, your call:**
- `reason` on reject is write-only — nothing reads it back. Surface it in the outgoing listing or drop the flag.
- `claim` never truncates inside `boundQueryPayload`; a 20k claim on the last surviving row overruns the 12k ceiling. `truncateText(claim, QUERY_EXCERPT_CHARS)` at :434 closes it.
- `generateId()` is the 8th copy of that one-liner across `src/store`. Pre-existing; a shared helper is a two-minute follow-up.
- Drop the 11 `"peer": true` lines from the lockfile — that's a different npm version, not a change.

**Scope, said plainly.** ~480 lines + 5 verbs + a tool + 2 tables + a migration for "a note about someone else's atom". What the smaller shape (an ordinary atom tagged with the target id) loses is real — revision-pinned staleness, owner-side reject/reopen, the guaranteed mark on the exact row rather than a competing search hit. I'm taking it because those are the properties that make it correct rather than merely present. But this is the last workspace feature of this size until something measures that people use the first one.

Merging after CI on 41468ab.

dat999zx:
Note: the fix commit was re-authored to the maintainer identity (4947ec3, same tree as 41468ab) so the CLA check passes.

dat999zx:
Rebased onto `main` after #284/#285/#286/#287/#292 landed. Two resolutions worth naming:

- **CHANGELOG** — both sides were additive `## Unreleased` entries; the dissent section sits above the OpenClaw fixes, nothing dropped.
- **`annotations` is now required on `ToolDefinition`** (#284), and the compiler caught `knowl_dissent` on the rebase — which is exactly what that field was added to do. Classified in df1b6e5 as **not** read-only: `list` reads, but `record`/`reject`/`withdraw`/`reopen` all write rows, and annotations are read before arguments, so a hint true for only some actions would lie. `destructiveHint` left at its default since every action is additive. Added to the `READ_ONLY` roster in `tool-annotations.test.ts` so the completeness check stays honest.

Gates on df1b6e5: build, typecheck, lint, docs:check clean; `tests/mcp` + `tests/workspace` + schema-pin + snapshot-table-ownership = 73 files, 573 tests, 0 failures.

Merging once CI is green.

PR #280, opened 2026-09-09
fix(openclaw): the session-start card must reach the model, and a workspace is a path, not a prefix

dat999zx:
Two OpenClaw plugin defects from the 2026-09-08 audit (§3.1 N1, N2). Spec and plan committed under `docs/superpowers/`.

**N2 — every OpenClaw session got zero engine memory.** The profile mapped `session_start` to `session-start`, which bound the session and composed the bootstrap card on a hook whose return value OpenClaw discards. The first `before_prompt_build` then found the session bound and emitted nothing. Measured: `session_start` → `before_prompt_build` gave `prependContext` length 0.

Fix follows the Hermes precedent (`src/session/hosts/hermes.ts:9`): `session_start` is no longer in the event map, so the first prompt hook binds the session and carries the card. The plugin still registers `session_start` for `warmWorkspace` only. The audit's suggested fix (assign the `withDeadline` result) was wrong — there is nowhere to deliver it.

**N1 — workspace matched by string prefix.** `cwd.startsWith(h.projectRoot)` let `knowl` capture `knowl-cloud`, so a hook in one repo read and wrote the other's store. Replaced with a `path.relative` boundary check at both call sites (`getHandle`, `releaseWorkspace`).

**Verification**
- Both new tests fail on the parent commit (rebuilt in a throwaway worktree) and pass after.
- Mutants: `isWithin → return true` fails the engine test; restoring `session_start` to the map fails both profile tests.
- No existing test modified. Full gate green (build, typecheck, lint, `npm test` 422/422).

Out of scope, per the audit's grouping: N3–N7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 Notable changes to `@dat999zx/knowl`. Versions before 2.1.0 predate this file; see the
 [git tags](https://github.com/dat999zx/knowl/tags) for that history.
 
+## Unreleased
+
+**The OpenClaw bootstrap card was being thrown away, on every session.** The profile registered `session_start`, so the engine bound the session and spent the bootstrap card on an event whose return value OpenClaw discards — and the first real turn (`before_prompt_build`) then arrived on a session the engine had already seen, with nothing to say. Measured: a fresh session whose first event is `before_prompt_build` gets the orientation card; the same session preceded by `session_start` got an empty answer with zero prepend context. That event is no longer mapped to the engine lifecycle, so the first `before_prompt_build` binds the session and carries the card. The plugin continues to register `session_start` solely to warm the workspace handle cache in memory so the initial write gate avoids cold open latency.
+
+**A workspace is a path, not a prefix.** `OpenClawEngineManager` matched cached workspace handles by string prefix (`cwd.startsWith(root)`), so sibling directories that share a name prefix (such as `knowl` and `knowl-cloud`) collided on lookup: whichever warmed first captured subsequent requests for the other, reading and writing atoms to the wrong project's database. Workspace matching in `getHandle` and `releaseWorkspace` now enforces a path boundary via `path.relative`, ensuring a workspace matches only itself or directories beneath it.
+
 ## 5.22.1 — 2026-09-07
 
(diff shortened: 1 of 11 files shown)


williamttruong:
We landed on the same two defects independently — I had #283 open with these plus three smaller ones before I saw this. Yours is the one to take; I'm closing mine and re-cutting only the parts this doesn't cover. Two things from my version that I think are real, both on `isWithin`:

**1. The path needs the Windows case fold, and this repo already has it.** `isWithin` runs `path.relative(root, cwd)` on raw values. `src/core/project-path.ts:16` exists for precisely this, and its docblock names the case:

> `path.resolve` alone is not enough on Windows. Its paths are case-insensitive but case-preserving, and the same project arrives with different casing depending on where the value came from: **a hook payload's `cwd` reports `D:\project` while `process.cwd()` in the hook process reports `d:\project`.**

A hook payload's `cwd` is exactly the input here. Unfolded, `D:\project` against a handle warmed at `d:\project` yields a relative path that climbs out, so the handle misses and the plugin opens the project a second time — the same split that docblock says "made an agent's own writes look foreign on its next tool event". Suggest `canonicalProjectRoot` on both sides before `path.relative`; it folds on win32 only, so POSIX case-sensitivity is preserved.

**2. Selection is still first-match-wins.** `Array.from(this.handles.values()).find(...)` returns whichever handle warmed first among those that contain `cwd`. With a monorepo at `C:/Code/mono` and a project of its own at `C:/Code/mono/packages/api`, both contain a file under the second, and the answer depends on open order rather than on ownership. Preferring the **longest** containing root is the same rule `findProjectRoot` applies walking up to the nearest marker.

Both are in #283 as `pathIsWithin` + `findCachedHandle` if the code is useful; steal freely, no attribution needed. The tests there include a Windows-gated case-fold case and a nested-root case, which are the two this would need.

One more from the same area, not in either PR and not worth blocking on: `disabledRoots` is written under `handle.projectRoot` but probed with the raw `cwd` (`engine.ts:190`), so a permanently-disabled workspace short-circuits only when the hook's cwd is exactly the project root. Correct, just pays a full `openProject` before refusing.

dat999zx:
Both review points taken, with one correction to the mechanism: `path.relative` already case-folds on win32 (`relative('D:\\Proj','d:\\proj\\src') === 'src'`), so `isWithin` itself was never the miss. The real gap was `handles`/`disabledRoots` keyed by the raw `handle.projectRoot`, so a second warm in different casing opened a duplicate handle. da268b9 keys both maps by `canonicalProjectRoot` and folds `isWithin` to match — which also closes the `disabledRoots` probe note. 9c07faf makes the longest containing root win. Both have tests that fail against the prior code.

The Windows CI failure (`claude-continuation-reminder.test.ts`, 0xC0000005 in the spawned CLI, empty streams) is a pre-existing flake — same file, same signature on main run 33945656223 from the 5.21.1 release push. Passes 3/3 locally on Windows on this branch. New push re-runs it.

PR #283, opened 2026-09-09
fix(openclaw): the plugin answers hooks from the right repo, and its session card reaches the model

williamttruong:
Four defects in the OpenClaw plugin from #260. It is the newest host integration and had not been audited; two of these mean an OpenClaw user's memory is either absent or someone else's.

Every fix was verified against a mutant rather than by reading the diff — for each one, the fix was reverted, the new test confirmed failing, then restored.

---

### 1. `fix(openclaw): a workspace is a path, not a prefix`

`engine.ts:131` was:

```js
const cached = Array.from(this.handles.values()).find((h) => cwd.startsWith(h.projectRoot));
```

No separator boundary, first-match-wins. On a machine holding `C:/Code/knowl` and `C:/Code/knowl-cloud` — an ordinary pair, since a project and its satellite share a prefix by convention — the second matches a handle warmed for the first. That handle carries query, the write gate **and** lifecycle capture, so a hook fired in one repository reads the other's atoms, is judged against the other's constraints, and writes capture rows into the other's store, with every layer reporting success.

New `pathIsWithin` uses `path.relative` for the boundary over `canonicalProjectRoot` — the fold the engine already applies to a path used as a key (`path.resolve`, plus a case fold **on Windows only**, because a hook payload's `cwd` reports `D:\project` where `process.cwd()` reports `d:\project`, while POSIX paths are genuinely case-sensitive). Selection now prefers the **longest** containing root rather than the first warmed, which is the same rule `findProjectRoot` applies walking up to the nearest marker — first-match is wrong for a monorepo that contains a project of its own.

### 2. `fix(openclaw): the session-start card must reach the model`

`index.ts:471` awaited `handle.lifecycle(normalized)` and never assigned the result. The card was composed and dropped — and the binding it created then forced every later `turn-start` down the `includeContext: false` branch, so OpenClaw users got **no engine memory at all**, silently.

Fixed in the plugin rather than in shared lifecycle code: the host-neutral context is parked and the first `before_prompt_build` of that session prepends it. Parked rather than recomputed because the pending handoff is consumed as the card is composed.

**Deliberately not the Hermes route.** Dropping `session_start` from the event map would also silently drop the session-start maintenance pass — abandoned-session recovery, event purge, read-set sweep, the auto-drift check and the two promotion passes — for OpenClaw users.

The `resolveWorkspace` docblock claimed a wrong guess "degrades to no memory this turn rather than to another project's database". That is false exactly when the gateway was started inside a repo. Behaviour left alone (removing the `process.cwd()` fallback would silence the plugin on every context carrying no directory); the docblock now says what actually happens.

### 3. `fix(openclaw): a card is keyed by a real session`

`sessionId` fell back to the literal `'openclaw-session'` on six paths, and `:310` deletes on read — so two concurrent sessions in one gateway shared one `midturnPending` slot and session B consumed session A's card. `impactSeen`'s `${sessionId}:${rel}` key collided the same way.

One helper replaces six copies of the fallback chain and returns `undefined` when OpenClaw named no session. It also stops treating `sessionKey` as an id: OpenClaw's contexts identify a run by `sessionKey` **and** `agentId` together, and the plugin's own tests carry two sessions both keyed `main`. Unattributable calls now drop the parked card and skip the impact cache rather than sharing a slot.

### 4. `fix(plugin): a library write is validated by the project's own settings`

`src/plugin.ts:93` called `storeKnowledgeItemDeduped(projectId, atom)` while the signature takes `(projectId, input, commitMessage?, validationOptions?)`. MCP passes `config?.security` and the CLI passes `config.security`; the plugin passed neither, so `secretPatterns` fell back to empty and a repo's own detectors never ran on a plugin write. Config is now read per write rather than captured at open.

Same commit: the migration-level guard at `engine.ts:175-177` caught, warned, and fell through to `handles.set` — so a database locked by a concurrent `knowl serve` (the normal case on Windows) was used anyway, possibly written by a newer knowl. It now releases and returns `null`, and deliberately does **not** add the root to `disabledRoots`, so a transient lock costs one warm attempt rather than the session.

---

## Verified

`npm run build`, `npm run typecheck`, `npm run lint`, `npm run docs:check` all clean. Full suite: **421 files passed, 3967 passed, 5 skipped, 0 failed.** `tests/integrations/openclaw/` on its own: 32 passed.

## Not changed — needs your call

**The write-gate matcher is still `['exec', 'apply_patch', 'spawn_agent']`** (`index.ts:278`), so no plain file-write tool name is matched. I could not settle it from a primary source: `openclaw` is a peer dependency deliberately not installed, so there are no vendored typings here.

The only in-repo evidence is second-hand and **contradicts what shipped**. `docs/superpowers/plans/2026-09-03-harness-integrations.md:888` records "Tool ids from docs.openclaw.ai/tools (2026-09-03): `read`, `write`, `edit`, `apply_patch`, `exec`" with `writesFiles: (_event, tool) => ['write', 'edit', 'apply_patch'].includes(tool)`. The shipped profile (`src/session/hosts/openclaw.ts:76`) and the plugin matcher both list `spawn_agent` — absent from that catalog — and omit `write` and `edit`.

If that catalog is current, the gate misses every ordinary file write and the fix is adding `write` and `edit` in both places. But the plan is a superseded draft rather than the docs, and changing a fail-closed gate on a superseded draft is not a call to make from here.

**One more, found while reading and left alone as out of scope:** `disabledRoots` is written under `handle.projectRoot` but probed with the raw `cwd` at `engine.ts:190`, so a permanently-disabled workspace short-circuits only when the hook's cwd is exactly the project root. Correct, just not cheap.

### DIFF
--- a/integrations/openclaw/README.md
+++ b/integrations/openclaw/README.md
@@ -89,6 +89,11 @@ run normally.
 
 Three of these carry constraints that are not obvious from the catalog:
 
+- **The session-start card also leaves through `before_prompt_build`.** OpenClaw ignores what a
+  `session_start` handler returns, so the card the engine composes there is parked in the plugin
+  and prepended to the first prompt of that session. It has to be: the session binding
+  `session_start` creates is what makes the engine's own turn-start answer empty, so a card left
+  unclaimed there is one nothing else ever produces.
 - **Recall never reads the prompt.** `before_prompt_build` emits a *fixed* orientation card. Building
   a query from the user's sentence is the defect fixed in knowl#257 on another host, and the
   "never prompts, never transcripts" promise depends on it not recurring here.
--- a/integrations/openclaw/src/engine.ts
+++ b/integrations/openclaw/src/engine.ts
@@ -1,5 +1,7 @@
+import path from 'node:path';
 import { createClient } from '@libsql/client';
 import {
+  canonicalProjectRoot,
   openProject,
   KNOWL_MIGRATION_LEVEL,
   type ProjectHandle,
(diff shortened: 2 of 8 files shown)


williamttruong:
Closing in favour of **#280**, which you opened for the same two headline defects (the path boundary and the session-start card) — we found them independently and yours is the one to take.

The two review points from this branch are on #280: the Windows case fold via `canonicalProjectRoot` (whose own docblock names the `D:\project` vs `d:\project` hook-payload case), and longest-match rather than first-match handle selection for nested roots. The code is in this branch's history as `pathIsWithin` and `findCachedHandle` if it is useful — steal freely.

The two defects this PR had that #280 does not touch are re-cut off `main` as **#285**: the plugin dropping the project's own `validationOptions` on every write, and the migration-level guard being fail-open on its own error. Full suite green there, 421 files / 3959 tests, guard mutation-verified after the re-cut.

The third, session-id keying (`'openclaw-session'` as a literal fallback on six paths, so two concurrent sessions share one `midturnPending` slot and `:310` deletes on read), is not in either PR — it is textually entangled with the card fix in #280. Happy to send it once #280 lands.

PR #284, opened 2026-09-09
feat(mcp): tool annotations, a discoverable resource template, and the asOf ceiling

williamttruong:
knowl pins `@modelcontextprotocol/sdk` at `^1.30.0`, which already ships tool annotations, resource templates, elicitation and completions — and registers exactly four request handlers (`ListTools`, `CallTool`, `ListResources`, `ReadResource`) declaring `capabilities: { tools: {}, resources: {} }` with no sub-flags. This takes the three pieces that are true today and leaves the rest alone.

Each change was verified against a mutant, not by reading the diff.

---

### 1. `feat(mcp): every tool declares what it does to memory before a host asks`

`annotations` is now a **required** field on `ToolDefinition`, typed as the SDK's own `ToolAnnotations` rather than a local copy of the same booleans — so the compiler asks a new tool the question instead of a reviewer having to. Policy comment at `tool-definitions.ts:17-58`.

Today every knowl tool looks equally dangerous to a host, so reading memory prompts as loudly as collecting it. Now: `readOnlyHint: true` on 16 tools, `destructiveHint: true` on `knowl_update`, `knowl_gc_apply` and `knowl_skill_run`, `openWorldHint: false` on all but `knowl_cloud`, `knowl_ingest` and `knowl_skill_run`, and a human `title` on 11 opaque names.

Three judgment calls worth your eye:

- **`knowl_drift` and `knowl_impact` are annotated as writes, not reads.** `apply` marks atoms for review (`tools.ts:1936`) and `resolve` closes a finding (`:1991-2006`), and annotations are read *before* arguments — so a hint that is only true for some argument values is a hint that lies.
- **Three read-only tools do incidental writes**, named in a source comment at `:33-42` so the claim is auditable rather than implied: `knowl_query` appends one row to the workspace demand ledger (best-effort, swallows its own failures, so not part of the contract); `knowl_transcript_search` catches up the index it is about to read; any embedding tool may populate the model cache on first use. None changes an answer a later call returns. If you consider any disqualifying, `knowl_query` is the one to revisit — it also carries most of this commit's value.
- **`idempotentHint` is on nothing**, deliberately. Every write mints an id, stamps a timestamp or appends an event, so repeating leaves the store observably different. The spec defaults it to false, which is the true answer. Equally, `destructiveHint` is left at its default (**true**, per `types.js:1190`) for `knowl_store`/`knowl_decide`/`knowl_ingest_atoms`, because the write engine can auto-retire a superseded predecessor — claiming `false` there would have been the easy lie.

The test is table-driven over every shipped tool, fails if a new tool arrives unclassified, and parses each tool through the SDK's `ToolSchema` — that schema **strips unknown keys**, so a misspelled hint would otherwise vanish silently rather than fail.

### 2. `fix(mcp): the category resource template is discoverable now, not just servable`

`resources.ts:99` already implements `knowl://category/{name}` and there was no `ListResourceTemplates` handler, so no host could discover it — working code, invisible. The handler returns the one template that actually resolves, with the seven valid names spelled out in the description (RFC 6570 cannot express an enum).

The test asks the running server and then **serves back everything it advertised**: an advertised template that 404s would be worse than the silence. Two existing tests that selected the read handler as `handlers[1]` by registration order now select by method name, which is why they show in the diff.

### 3. `fix(mcp): a historical knowl_query answers under the same ceiling as a live one`

`tools.ts:904` returned `compactMcpJson(items.map(compactItemResponse))` on the `asOf` branch with no `boundQueryPayload`, bypassing `MAX_RESPONSE_CHARS = 12_000` entirely. Measured on this repo's fixture: **32,451 characters against a 12,000 ceiling.** Both paths now emit an identical bounded notice.

### 4. `test(mcp): the capability card is pinned to what the server actually answers`

**No capability was added, because none would be true.** The test exact-compares the declared card, asks every method it promises and expects an answer, then asks `resources/subscribe` and `prompts/list` and expects `-32601`. A flag cannot be added without the mechanism, nor the mechanism without the flag. `server.ts:73-90` documents why the literal stays as it is.

---

## Sources verified at primary

- `annotations` sits **on** the Tool object — `@modelcontextprotocol/sdk/dist/esm/types.js:1263`, field shapes at `:1173-1211`.
- `openWorldHint: false` is the spec's own wording for this case — `types.js:1202-1208`: *"the world of a web search tool is open, whereas that of a memory tool is not"*.
- `resources/templates/list` is gated on `capabilities.resources` alone (`server/index.js:221-226`), so there is no template sub-flag to declare.

**`anthropic/maxResultSizeChars` — verified real, and deliberately NOT added.** Per code.claude.com/docs/en/mcp it *raises* a tool's persist-to-disk threshold up to 500,000 characters; it is not a hint that a tool returns small results. `knowl_query` is already bounded at 12,000, below the default threshold, so emitting it would either do nothing or request a budget knowl does not want. The original framing of this was inverted.

## Verified

`npm run build` · `npm run typecheck` · `npm run lint` · `npm run docs:check` all clean. Full suite: **424 files, 3970 passed, 5 skipped, 0 failed.**

Live over stdio against `node dist/index.js serve`: 29 tools listed, 29 annotated; `knowl_query` → `{"readOnlyHint":true,"openWorldHint":false}`; `knowl_gc_apply` → `{"title":"Delete retired memory","destructiveHint":true,"openWorldHint":false}`; the template lists; `resources/subscribe` → `-32601`. Independently re-derived here: 28 tools under a null config, **zero tools carrying both `readOnlyHint: true` and `destructiveHint: true`**.

## Note on overlap

`fix/audit-store` (opened alongside) also touches `tools.ts` and `tool-definitions.ts` for the GC preview/apply contract. No semantic conflict, but merge order will decide which one rebases.

### DIFF
--- a/src/mcp/resources.ts
+++ b/src/mcp/resources.ts
@@ -1,6 +1,6 @@
 import { Server } from '@modelcontextprotocol/sdk/server/index.js';
-import { ErrorCode, ListResourcesRequestSchema, McpError, ReadResourceRequestSchema } from '@modelcontextprotocol/sdk/types.js';
-import { KnowledgeCategory } from '../core/types.js';
+import { ErrorCode, ListResourceTemplatesRequestSchema, ListResourcesRequestSchema, McpError, ReadResourceRequestSchema } from '@modelcontextprotocol/sdk/types.js';
+import { KNOWLEDGE_CATEGORIES, KnowledgeCategory } from '../core/types.js';
 import { getRecentContext } from '../store/recent-context.js';
 import { formatRecentContextToMarkdown, formatHierarchyToMarkdown } from '../core/format.js';
 import { getHierarchicalKnowledge, queryKnowledgeBase } from '../store/queries.js';
@@ -38,7 +38,35 @@ export function registerResources(
     };
   });
 
-  // 2. Read resource
+  // 2. List resource templates
+  //
+  // `knowl://category/{name}` has been resolvable by the read handler below since it was
+  // written, and no client could ever find out: `resources/templates/list` is the ONLY place
+  // the protocol lets a server advertise a parameterised URI, and nothing answered it. A host
+  // asking got the SDK's "method not found", so working code was invisible -- not gated, not
+  // deprecated, simply undiscoverable.
+  //
+  // One entry, because one template resolves. The read handler matches `[a-z]+` and then hands
+  // whatever it caught to `queryKnowledgeBase`, so a name outside the category set answers with
+  // an empty document rather than an error -- which is why the valid names are spelled out in
+  // the description. RFC 6570 has no way to say "one of these seven", and the description is
+  // where a client's model reads what to substitute.
+  server.setRequestHandler(ListResourceTemplatesRequestSchema, async () => {
+    return {
+      resourceTemplates: [
+        {
+          uriTemplate: 'knowl://category/{name}',
+          name: 'category',
+          title: 'Active Knowledge by Category',
+          description: 'Every active item in one category, as markdown. `name` is one of: '
+            + `${KNOWLEDGE_CATEGORIES.join(', ')}.`,
+          mimeType: 'text/markdown',
+        },
+      ],
+    };
+  });
+
+  // 3. Read resource
   server.setRequestHandler(ReadResourceRequestSchema, async (request) => {
     const { uri } = request.params;
     await whenReady();
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -68,6 +68,25 @@ export function createMcpServer(
       version: PACKAGE_VERSION,
     },
     {
+      // Exactly what is implemented, and nothing that merely could be.
+      //
+      // A host trusts this card: it decides from these flags whether to subscribe, whether to
+      // re-list after a change, and what to stop asking for. A capability declared and not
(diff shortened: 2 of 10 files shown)


dat999zx (approved):
All four claims hold against the code. Every read-only-marked handler grepped for writes; the wire path is direct (`knowlToolDefinitions()` objects unchanged into `ListTools`); mutant `knowl_query readOnlyHint→false` fails two tests. `asOf` now shares `boundQueryPayload` + `responseBoundedNotice` with the live path, so the notice shape is identical by construction. The capability pin is not SDK-brittle: `getCapabilities()` returns the options object verbatim and `-32601` is what the SDK emits for any unregistered method.

One doc inaccuracy fixed and pushed as b34655c: the rationale comment said the only incidental write on read-only tools is the demand-ledger row. Live `knowl_query` and `knowl_context` go through `queryKnowledgeForAgentExplained`, which inserts one `knowledge_access` row per result, and `gc.ts:55 isHot` reads those — so `knowl_gc_preview` answers differently after heavy querying. `readOnlyHint: true` is still right (telemetry, not memory); the comment now says so.

`knowl_cloud openWorldHint: true` is about domain, not I/O — no action there makes a network call. Fine as is.

PR #285, opened 2026-09-09
fix(plugin): a library write is validated by the project's own settings, and an unverifiable database is not opened

williamttruong:
Complementary to **#280** — no overlap. That PR takes the path boundary and the session-start card; this takes two defects in the same plugin that it does not touch. I originally had all four in #283 before I saw #280; that PR is closed and this is the remainder, re-cut off `main`.

Verified against a mutant rather than by reading the diff.

---

### `handle.store()` dropped the project's own security settings

`src/plugin.ts:93` called:

```js
storeKnowledgeItemDeduped(projectId, atom)
```

while the signature is `(projectId, input, commitMessage?, validationOptions?)`. The MCP surface passes `config?.security` (`src/mcp/tools.ts:650`) and the CLI passes `config.security` (`src/cli/program.ts:2803`); the plugin passed neither, so `secretPatterns` fell back to the empty list.

Two consequences, in opposite directions: a repository that configured extra secret detectors got **default** validation on every plugin write, and one that relaxed them got writes **refused** that its own config allows. Config is now read per write rather than captured at open, so a change to `.knowl/config.json` takes effect on the next write instead of on the next gateway restart.

### The migration-level guard was fail-open on its own error

`engine.ts` caught, warned, and fell through to `handles.set(root, handle)`. So the one case the guard exists for — the level cannot be read — was the case it let through. On Windows the ordinary cause is a concurrent `knowl serve` holding the file, which makes that the common path rather than the rare one.

It now releases the handle and returns `null`. An older plugin writing into a store a newer knowl migrated finds every table it expects, because the schema is `CREATE TABLE IF NOT EXISTS` plus additive `ALTER`s, and then writes rows the newer schema's invariants do not hold for — with nothing reporting it.

**The root is deliberately not added to `disabledRoots`.** A lock clears. A transient failure costs this warm attempt and the next hook in that workspace retries; the sticky list is for the one condition that cannot resolve itself, a database stamped past what the plugin understands.

---

## Verified

`npm run build`, `npm run typecheck`, `npm run lint` all clean. Full suite: **421 files passed, 3959 passed, 5 skipped, 0 failed.** Scoped: `tests/integrations/openclaw/` + `tests/cli/plugin-export.test.ts` → 31 passed.

**Mutant check on the guard, run here after the re-cut** (since re-cutting changed the test file): removing the `release()` + `return null` pair makes exactly the three guard tests fail — `refuses the workspace and releases the handle…`, `does not cache the unverified handle…`, `is not sticky…` — and restoring it returns a clean tree with no diff.

The guard tests stub `openProject` on the module object, so the test file imports `@dat999zx/knowl/plugin` namespaced as well as named; that import is the only line here that is not from the original commit.

## Note

`tests/integrations/openclaw/engine.test.ts` grows one `describe`. The path-boundary tests that were in #283 are gone from this branch — they belong with #280's fix, and I have left the two review points on that PR instead.

### DIFF
--- a/integrations/openclaw/src/engine.ts
+++ b/integrations/openclaw/src/engine.ts
@@ -173,7 +173,25 @@ export class OpenClawEngineManager {
         return null;
       }
     } catch (err: unknown) {
-      this.logger?.warn?.(`[knowl] Could not verify migration level for ${handle.databasePath}: ${err}`);
+      // Fail CLOSED, and not sticky.
+      //
+      // This branch means the level could not be READ -- on Windows a concurrent `knowl serve`
+      // holding the file is the ordinary cause -- so the one thing actually known is that the
+      // database is unverified. Warning and continuing to `handles.set` defeated the check
+      // entirely: an older plugin writing into a store a newer Knowl migrated finds every table
+      // it expects, because the schema is `CREATE TABLE IF NOT EXISTS` plus additive `ALTER`s,
+      // and then writes rows the newer schema's invariants do not hold for. Nothing reports it.
+      //
+      // The root is deliberately NOT added to `disabledRoots`: a lock clears. A transient
+      // failure costs this warm attempt, and the next hook in that workspace tries again --
+      // where the sticky list is for the one condition that cannot resolve itself, a database
+      // stamped past what this plugin understands.
+      this.logger?.warn?.(
+        `[knowl] Could not verify the migration level of "${handle.databasePath}", so this workspace ` +
+        `is not being opened. Knowl will try again on the next event: ${err}`,
+      );
+      await safely(() => handle.release(), this.logger);
+      return null;
     }
 
     this.handles.set(root, handle);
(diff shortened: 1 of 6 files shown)


dat999zx (approved):
Guard half: correct, mutant-verified (restoring the fall-through fails all three guard tests), and every consumer of a null handle in `index.ts` treats it as "no memory this turn" — nothing falls back to the global store. Not adding to `disabledRoots` is right.

Validation half: right defect, wrong layer. The plugin was one of **three** callers omitting `validationOptions` — `indexSkillPackage` (knowledge-index.ts:16) and `promoteSessionCandidates` (candidate-promotion.ts:35) have the same fall-through, and the promotion path is the one most likely to carry a pasted credential. And the per-write `loadConfig` is the exact call `knowledge-writer.ts:636-641` removed after it crashed a 2500-write run; `.catch(() => null)` also degrades to `secretPatterns=[]`, weaker than `DEFAULT_CONFIG`.

Pushed 61a4208: the default moves into `storeKnowledgeItemDeduped` / `storeKnowledgeAtomsDeduped` with a per-root cache beside `workspaceCache`, falling back to `DEFAULT_CONFIG.security` on a read failure. Your plugin change reverts to the one-liner; your test passes unchanged, now carried by the writer. Three new tests in `tests/store/write-security-default.test.ts`; two fail when the default line is removed.

PR #286, opened 2026-09-09
fix(store): five defects — secret scanning, a bootstrap rollback, import conflicts, an atomic update, and unpreviewed purges

williamttruong:
Five store defects from an audit of `main`. One revertable commit each, every one verified against a mutant rather than by reading the diff.

**No new table, no new migration level** — still 16, so this does not collide with #279's 17.

---

### 1. `fix(store): every write path scans every field it writes`

`validateKnowledgeWrite` reads `tags` and `alternatives`, but three callers handed it a five-field object literal and an absent field makes `arrayField` return `[]` — so the scan covered nothing:

- `src/store/integrity.ts:118-137` selects and parses both arrays, then drops them. So `knowl audit` and `knowl doctor` print **"No integrity findings" over a store whose `tags` column holds a live credential** — and snapshot-restore verification inherits the false clean. A check that answers "clean" is worse than no check.
- `src/store/repository.ts:363-370` validates five fields and writes the whole spread at `:425`. Creating an atom with a credential in `tags` is refused; creating it clean and then *updating* with the same credential is accepted.
- `src/store/portability.ts:657` — same hole on the path that ingests a peer's export.

All three now use one projection (`scannableFields`) taken off the record, which copies only the keys the caller supplied — so the update path scans exactly what it writes and a metadata-only edit does not re-scan stored prose.

**Scope note, so this is not oversold:** this is local-store integrity, snapshot trust and import trust. It is *not* a disclosure bug — knowl-cloud's publish path already scans tags and alternatives before authorization, so a poisoned tag wedges the outbound queue rather than reaching a team store.

Mutant: reverted the five files → 4 of 5 tests fail; the fifth is the false-positive guard that must pass both ways.

### 2. `fix(store): a skill FK repair must not roll back bootstrap` — shipped alone

`repairSkillForeignKeys` copies every row into a table whose new FK is `ON DELETE CASCADE`. Its own `PRAGMA foreign_keys = OFF` is inert inside the migration transaction — `bootstrap.ts:1278-1280` already says so — and `defer_foreign_keys` pushes the check to COMMIT, where a failure rolls back the **entire bootstrap**: `application_id` stays 0, ~40 tables are never created, and every later open fails identically. `doctor` calls `initDb` too, so the repair path is inside the failure.

**The finding's stated trigger is refuted, and the commit body carries both halves.** "Any store with an orphan `skill_steps` row" does *not* brick: `DROP TABLE` decrements the deferred-FK counter once per violating row it deletes, so when the stale key points at a table that no longer exists each orphan is +1 on the insert and −1 on the drop, and COMMIT survives. Verified three ways — a standalone libSQL probe of the exact rename/create/insert/drop sequence (`nodrop` throws, `full` commits), a variant with an extra unbalanced orphan (throws, so the drop decrements rather than resets), and instrumented `bootstrapSchema` runs.

**The brick is still reachable** — when the stale key points at a table that still exists and satisfies the row, nothing cancels — and that is the case the first test pins. Both cases leave a row the new key forbids (`PRAGMA foreign_key_check` reports it); only one announces itself. Say the word if you want the commit reframed as *"the repair leaves a row its own key forbids"* rather than *"bricks the store"*; the body already supports either.

Mutant: reverted `bootstrap.ts` → both tests fail, the live-target case with `SQLITE_CONSTRAINT` thrown from the COMMIT at `:1304`.

### 3. `fix(store): an import cannot land a second exclusive value`

`portability.ts:769` was a raw INSERT with no `checkKnowledgeConflict` caller in the file. A direct second exclusive write is refused with `KNOWLEDGE_CONFLICT`; importing a peer's export landed it and reported `conflicts: 0`.

Import now resolves every planned write against the identities the store holds — normalised on both sides, since a raw stored key would never collide with a normalised one — counts a collision in `conflicts` (which declines the whole import and writes nothing, the shape `--on-divergence fail` already has), and names the pairs. A holder the same import retires, archives or deletes releases its identity first, so an export *replacing* a value is not refused for colliding with the value it replaces.

Mutant: 2 of 5 fail — the two asserting the refusal; the other three are false-positive guards.

### 4. `fix(store): an update and its commit are one transaction`

`knowledge-actions.ts:217` ran the row and the change record as two transactions on two connections; the comment at `:246` states the problem. A throwing `createKnowledgeCommit` left the item `superseded` with `supersededById` set and nothing logging it — invisible to the change notice, blast radius and `readCommitHead`, while the row says it happened and the caller is told it failed.

Both now run inside one `withClientTransaction`, following `storeKnowledgeItemDeduped`. The three best-effort side effects (sibling flagging, re-embed, cloud staging) stay outside and after, so nothing stages a row a rollback removed.

Mutant: `expected 'superseded' to be 'active'` — exactly the described leak.

### 5. `fix(store): collection acts on the set the preview named`

`gc.ts:319-321` recomputed candidates and the MCP tool took no arguments, so: preview shows `purge: 0`, an atom is written, apply hard-deletes an item the preview never named. Purge is the one GC action with no undo.

**The contract choice is yours to confirm.** The gate is on `purge` alone. `knowl_gc_apply` gains optional `purgeItemIds`, `knowl gc` gains `--purge <ids>`, and an apply naming nothing purges nothing while reporting what it declined (`unapprovedPurges`). Archive and compress still run unapproved — both leave the item, its id, its assertions and its evidence links in place, so a wrong one is recoverable, and gating them would leave an argument-less apply doing nothing at all. That is what makes a no-preview apply *safe* rather than merely refused. Apply still recomputes, because a stale plan is its own bug; it purges only the intersection with the approved ids.

Ids are carried by the caller rather than persisted — `knowl gc` and `knowl gc --apply` are separate processes, so no table and no migration level. `knowl_gc_preview` now returns every purge id, since it is the response that must supply them.

Mutant: *"destroys nothing an empty preview could not have named"* fails with `expected 1 to be +0` — the unpreviewed atom was deleted.

---

## Verified

`npm run build`, `npm run typecheck`, `npm run lint` (0 errors, 0 warnings), `npm run docs:check` all clean. Full suite: **426 files passed, 3973 passed, 5 skipped, 0 failed.** The schema-pin and snapshot-table-ownership guards passed untouched.

## Notes

- **Five existing test files changed** for the commit-5 contract (`forget-log`, `gc-duplicate-survivor`, `gc-protections`, `store`, `mcp/server`). Each now previews and approves, which is what a caller does; no assertion was weakened.
- **Not done, flagged rather than scoped in:** `repository.ts` has a separate `steps` parameter that is a genuine unscanned field on the create/update path. It sits outside the three call sites above and carries real refusal risk for existing skills, so it is named here rather than fixed.
- **Overlap:** #284 also touches `tools.ts` and `tool-definitions.ts` (tool annotations). No semantic conflict, but merge order decides which rebases.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -2020,7 +2020,12 @@ lands. Best-effort vector indexing after import uses only the locally available 
 ```bash
 knowl gc
 knowl gc --apply
+knowl gc --apply --purge <id,id,...>
 ```
+
+Apply archives and compresses on its own, but purges nothing unless `--purge` names ids the
+preview listed; only ids that are still candidates are deleted, and the rest are reported.
+`knowl_gc_apply` takes the same approval as `purgeItemIds`, which `knowl_gc_preview` returns.
 
 The default policy:
 
@@ -2719,7 +2724,7 @@ knowl eval --dataset docs/evals/retrieval-suite.json --json
 | `knowl park --goal <goal> [--next-action <a>] [--completed <c...>] [--blocker <b>] [--artifact <p...>] [--verified\|--unverified]` | Park a workstream and get a key back; hand the key to the user verbatim |
 | `knowl handoff --goal <goal> --next-action <a> [--completed <c...>] [--blocker <b>] [--artifact <p...>] [--verified\|--unverified]` | Leave a baton the next session in this project receives once |
 | `knowl resume [key]` | Resume a parked workstream from its key, or list what is parked here |
-| `knowl gc [--apply] [--stale-days N] [--compress-days N] [--min-bytes N] [--ignore-access] [--tombstone-days N]` | Preview or apply duplicate, archive, compression, and tombstone maintenance |
+| `knowl gc [--apply] [--purge <ids>] [--stale-days N] [--compress-days N] [--min-bytes N] [--ignore-access] [--tombstone-days N]` | Preview or apply duplicate, archive, compression, and tombstone maintenance |
 | `knowl forget-log [--limit N] [--repo <name>] [--json] [--prune-days N]` | Show why knowledge items were destroyed — policy, reason, and the retrieval evidence it overruled — or prune those records |
 | `knowl pr --since <commit> [--dry-run]` | Find drift candidates and, unless dry-run, mark them for review |
 | `knowl reviewed <itemId> [--note <text>]` | Record that an item was re-read and still holds, clearing its review flag here and on the team's copy |
--- a/src/cli/program.ts
+++ b/src/cli/program.ts
@@ -2450,6 +2450,16 @@ program.command('import').argument('<path>').description('Load portable JSONL me
           `Import did not apply${result.conflicts ? ` (${result.conflicts} conflict(s))` : ''}. ` +
           'Re-run with a different --on-divergence policy, or reconcile the source.',
         );
+        // No divergence policy resolves an exclusive collision: two items both claim to be
+        // the one active value for a key, and only a person can say which. Naming them is the
+        // difference between "try another policy" -- which cannot work -- and an actionable
+        // instruction.
+        for (const clash of result.exclusiveConflicts ?? []) {
+          console.error(
+            `  ${clash.id} claims an exclusive identity already held by the active item ` +
+            `${clash.heldBy}. Retire one of them, then re-import.`,
+          );
+        }
         process.exitCode = 1;
       }
     } catch (error: any) {
--- a/src/cli/program.ts
+++ b/src/cli/program.ts
@@ -3717,6 +3717,10 @@ program
   .option('--min-bytes <bytes>', 'Minimum content bytes before compressing an archived item (default 180)')
   .option('--ignore-access', 'Archive stale state even if it was recently or frequently retrieved (hot)')
   .option('--tombstone-days <days>', 'Remove delete records older than this many days (default 90)')
+  // Purge is the one action with no undo, and `knowl gc` and `knowl gc --apply` are two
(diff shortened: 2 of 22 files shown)


dat999zx (approved):
Five for five, each verified against the code path it names, each mutant killed. Notes per commit:

1. `scannableFields` earns its keep only on the update path; audit and import could hand the record straight to `validateKnowledgeWrite` (it ignores unknown keys). Not worth churning. The `steps` hole you flagged is closed and pushed as 4836403 — joined onto the scanned input on create and update, `stringFields` accepts bare-string steps, test added.
2. Real fix, not a failure-mode swap: the `WHERE knowledge_item_id IN (SELECT id FROM knowledge_items)` filter means no deferred violation is ever created. The removed pragmas were dead. Reverting the filter fails both tests with `SQLITE_CONSTRAINT` at COMMIT.
3. Same normaliser as `checkKnowledgeConflict`; self-import guarded by `holder !== item.id` plus the identical/keep-local skip; holders retired by the same file release first.
4. Lock window is strictly *shorter* than before — git spawn and `getKnowledgeItem` run before the tx, flagging/re-embed/staging after.
5. Only two callers of `applyKnowledgeGc` and both are updated; no automated path purges. But `knowl gc --apply` without `--purge` now purges nothing, and someone's cron may run exactly that. The CLI says so on stdout, which is enough — flag it in the release notes when this ships. `--purge` / `purgeItemIds` were undocumented in `docs/reference.md`; fixed in 4836403.

dat999zx:
Note: the fix commit was re-authored to the maintainer identity (ae94184, same tree as 4836403) so the CLA check passes.

PR #287, opened 2026-09-09
fix(cli): five defects — a global-store fallback, two hook-lifecycle asymmetries, unfailable doctor checks, and inert options

williamttruong:
Five defects across the CLI and the session-hook layer. One revertable commit each, every one verified against a mutant.

---

### 1. `fix(cli): a broken project store is not a global-store question`

`program.ts:812-820`. One `try` spans `findProjectRoot`, `initDb` and the project-row lookup, and the `catch` never inspected `err` — it gated only on `globalOnlyNamespaces().length > 0`. Only the *first* failure means "no project here", so a repository whose store will not open was answered from the machine-wide store.

Reproduced in a throwaway temp project under a `KNOWL_HOME` override:

```
knowl status  → ❌ Error reading status: … SQLITE_NOTADB: file is not a database   exit=1
knowl list    → exit=1
knowl query "entry point app.js" → {"title": "GLOBAL CANARY", …}                   exit=0
```

`status` and `list` report the real error; `query` returns another store's atom as this project's answer and exits 0. Now gated on `shouldUseGlobalStore`, the predicate `resolveCloudTarget` already uses — the CLI-side twin of the MCP server's `shouldServeGlobalOnly`. A fourth call site, not a fourth copy.

Mutant (guard broadened back to the bare check): both tests fail, one on the canary text, one on `expected +0 to be 1`.

### 2. `fix(hooks): a failed write is not a durable write`

`host-lifecycle.ts:1150`. No status guard, while the turn counter fifteen lines below already has `input.status !== 'failed'` and the comment at `:1163` states the opposite principle. So a refused or thrown `knowl_store` counted as a durable write **and settled pending lessons** — the session is marked as having captured knowledge it did not capture.

**The fixture was also de-flaked in the same commit, and it matters.** Roots were `.knowl-capture-nudge-<counter>` with the counter restarting at 1 each run, while `afterAll`'s `fs.rm` is exactly the removal Windows refuses while libSQL holds the `-shm` sidecar. Four leftovers were found; each handed test N a store carrying a previous run's counters, and the file failed a different 2–3 assertions per run. Roots now carry a per-run token, verified over three clean consecutive runs before and after the mutant check.

### 3. `fix(hooks): enforce records at least what shadow records`

`host-lifecycle.ts:819-828`. Shadow claims the nudge; enforce hits `if (!profile.stopContext) return undefined` and returns first — the stricter mode observing *less* than the looser one. The undeliverable-host branch now claims `shadow`, staying distinct from `enforce` so the capability check's property holds.

The existing test asserted `nudged: null` there, which encoded the defect; it now asserts `'shadow'`, plus a new monotonicity test. Mutant: both fail.

### 4. `fix(cli): doctor checks the surface it actually serves`

`doctor-report.ts:205-229, 303-305`. Three checks tested `KNOWL_MCP_TOOL_NAMES` — an `as const` flattened from prose routing groups. Unfailable. A repository with **zero** integrations printed "[OK] MCP tools expose knowl_query", "[OK] …task tools" and "[OK] …skill bridge tools" directly beside "[OK] No agent MCP integration selected".

Now `knowlToolDefinitions(config)`, dynamically imported — `mcp/tools.js` otherwise drags the MCP SDK, the AI pipeline and the embedder into every CLI command. Mutant (revert to the constant): all four subtraction cases report OK and fail.

### 5. `fix(cli): options that are read, and errors that are caught`

The cosmetic batch, three findings in one commit.

**C2 reproduced:** `knowl agent-reminder` → `TypeError: Cannot read properties of undefined (reading 'charAt')` plus a stack, because `reminder.ts:113-114` sit outside the try and `src/index.ts:55-61` dispatches past commander, leaving its error handler dead and `<host>` unenforced. The host is validated now, the emission lines have a try of their own, and a new `reportHookFailure` owns `refusesOnAnyNonZeroExit` for **both** commands — the reminder had no equivalent and runs on Copilot, where a non-zero exit refuses the turn.

**C4:** both `--budget` sites and `Number(options.expiresIn)`. `--budget abc` produced NaN, `Date.now() >= NaN` is false forever, and the paid model ran unbounded.

**C5 — implemented rather than deleted.** `--json` is already written into every installed hook config, so churning it out of them is the more disruptive option. It now carries the meaning `reportCommandFailure` already gives it on the sibling `agent-lifecycle`: failure also writes a parseable envelope to stdout, because a hook parses stdout. Host lists in both help texts are now derived from `HOST_PROFILES` — `hermes` and `openclaw` were missing, and both emit cards.

Test spawns the real CLI and reads real streams, deliberately avoiding the `process.stdout.write`-spy trap (under vitest that spy does not catch `console.log`, so such a test passes against its own mutant). Mutant (`git stash push -- src/`, rebuild): **7 of 8 fail** — `expected '…dist/rem…' not to contain 'TypeError'`, two `SyntaxError: Unexpected end of JSON input` from the inert `--json`, `expected 1 to be +0` from `--json` being taken as the host, and both coercion tests. The 8th is the fractional-budget guard, which correctly passes both ways.

**One deliberate deviation from the brief:** `--budget` uses a new `positiveNumber` beside `positiveInt` rather than `positiveInt` itself. `<minutes>` is documented as minutes and `positiveInt` would refuse `--budget 0.5` — a valid input — to catch an invalid one. A test asserts `0.5` still passes coercion. One-word change if you disagree.

---

## Verified

`npm run build`, `npm run typecheck`, `npm run lint` (0 problems), `npm run docs:check` ("Generated documentation regions are current") all clean; `docs:generate` produced no drift, so the help-text changes needed no regeneration. Full suite: **423 files passed, 3973 passed, 5 skipped, 0 failed.**

## Not fixed — a sibling worth your call

`evaluatePendingLessonStop` (`host-lifecycle.ts:869`) carries the identical `if (!profile.stopContext) return undefined` that commit 3 fixes, where shadow mode would have marked the open lessons `'shadow'`. Same shape, same backwards monotonicity — but it is a new finding rather than part of that fix, so it is named here rather than folded in.

### DIFF
--- a/src/cli/agent-hook.ts
+++ b/src/cli/agent-hook.ts
@@ -6,7 +6,8 @@ import { handleHostLifecycleEvent } from '../session/host-lifecycle.js';
 import { assertKnowledgeDatabasePresent } from './database-presence.js';
 import { readLifecyclePayload } from './agents/lifecycle.js';
 import { IncompleteHostHookPayloadError, normalizeHostHook } from './agents/host-hook.js';
-import { hostProfile, isHookHost } from '../session/hosts/index.js';
+import { hostProfile } from '../session/hosts/index.js';
+import { HookOutputOptions, reportHookFailure } from './agents/hook-failure.js';
 
 /**
  * The agent lifecycle hook, in its own module and off the CLI's import graph.
@@ -26,7 +27,7 @@ import { hostProfile, isHookHost } from '../session/hosts/index.js';
  * inside the one branch that uses it, because it reaches the embedding provider and only
  * turn-stop and session-stop events ever get there.
  */
-export async function runAgentHook(host: string, event: string): Promise<void> {
+export async function runAgentHook(host: string, event: string, options: HookOutputOptions = {}): Promise<void> {
   try {
     const payload = await readLifecyclePayload();
     const normalized = normalizeHostHook(host, event, payload);
@@ -96,11 +97,15 @@ export async function runAgentHook(host: string, event: string): Promise<void> {
       await closeDb().catch(() => {});
       return;
     }
-    console.error(`Error handling agent hook: ${error.message}`);
     await closeDb().catch(() => {});
-    // See `refusesOnAnyNonZeroExit`: on those hosts this exit would deny the agent's edit
-    // rather than report our own crash. The error still reaches stderr; only the status goes.
-    if (isHookHost(host) && hostProfile(host).refusesOnAnyNonZeroExit) return;
-    process.exit(1);
+    // Both streams under `--json`, and the `refusesOnAnyNonZeroExit` rule, now live in
+    // `reportHookFailure` so the reminder obeys the same two. On those hosts a non-zero exit
+    // would deny the agent's edit rather than report our own crash, so the error still reaches
+    // stderr and only the status goes.
+    //
+    // `exitCode`, not `exit`: this path now writes an envelope to stdout first, and `process.exit`
+    // can truncate a pipe mid-write. The store is already closed, so nothing holds the loop -- as
+    // the refusing-host branch has relied on since it was added.
+    if (reportHookFailure(host, options, 'Error handling agent hook', error)) process.exitCode = 1;
   }
 }
--- a/src/cli/agents/hook-failure.ts
+++ b/src/cli/agents/hook-failure.ts
@@ -1,0 +1,42 @@
+import { HookHost } from '../../core/host-hook-types.js';
+import { HOST_PROFILES, hostProfile, isHookHost } from '../../session/hosts/index.js';
+
+/** What `--json` means on a hook command, in the one shape both of them take. */
+export type HookOutputOptions = { json?: boolean };
+
+/** Every host declaring a prompt event, which is exactly the set `agent-reminder` can serve. */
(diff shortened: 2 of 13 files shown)


dat999zx (approved):
Five real defects, five root-cause fixes, each mutant killed. `shouldUseGlobalStore` can never return true for a corrupt-but-present store (`ProjectNotFoundError` is thrown only by `findProjectRoot` on a missing `.knowl`); and `program.ts:829` was the only ungated `globalOnlyNamespaces().length` site, so there are no siblings left. `status` is `'finished' | 'failed' | undefined` — the guard is complete.

The sibling you named: `evaluatePendingLessonStop` had the identical hole. Fixed and pushed as b5b47c2 with a test that fails against the original.

One thing 92ebd8d makes visible rather than causes: `status-report.ts:274` labels the count "Nudges delivered" whenever mode is `enforce`, and `captureHealth` counts any non-null `nudged`. On a non-Claude host under enforce, `knowl status` now reports would-haves as delivered. Was already mislabelled after a mode flip; now wrong on a fresh repo. Count `'enforce'` and `'shadow'` separately — separate small PR, not this one.

Nits, take or leave: `REMINDER_HOSTS`/`HOOK_HOSTS` are host-registry facts and belong in `session/hosts/index.ts`, not `hook-failure.ts`; `reportCommandFailure` could wrap `reportHookFailure` to drop four duplicated lines.

Issue #290, opened 2026-09-09
No host re-delivers the bootstrap card after compaction: every compaction event maps to checkpoint


dat999zx:
## Summary

Every host's compaction event normalises to `checkpoint`, and the session-start branch never distinguishes a post-compaction restart from a resumed session. After the host compacts, the session runs with no Knowl card — and on hosts where the rules ride the card rather than `CLAUDE.md`/`AGENTS.md`, with no rules — until the next real session start.

## Evidence

Compaction → `checkpoint` on every host that has one:
- `src/session/hosts/claude.ts:11` `PreCompact`
- `src/session/hosts/copilot.ts:26` `preCompact`
- `src/session/hosts/cursor.ts:13`
- `src/session/hosts/hermes.ts:36` `on_pre_compress`
- `src/session/hosts/openclaw.ts:36` `before_compaction`
- `src/session/hosts/codex.ts:25-26` deliberately leaves `PostCompact` unmapped

The checkpoint branch (`src/session/host-lifecycle.ts:1131-1144`) captures an event and composes no context. The session-start branch (`:947-1026`) never reads `source`, and `source` is not in `ROOT_FIELDS` (`src/cli/agents/lifecycle.ts:13-37`) so it could not. `bootstrapWithHandoff` on an already-bound session takes the existing-session path in `getOrCreateHostSession` (`src/session/host-session-bindings.ts:53-63`) with `includeContext: false`.

antigravity, openhands and cline have no compaction event at all.

## Reproduction (on `dist/index.js`)

1. `knowl init --yes` in a throwaway dir, store one atom
2. `agent-hook claude SessionStart` with `session_id: c-1` → full card
3. `agent-hook claude SessionStart` with `session_id: c-1, source: 'compact'` → **roster only, no knowledge card**
4. `agent-hook claude UserPromptSubmit` with `session_id: c-1` → empty

Same on hermes: `on_pre_compress` then `pre_llm_call` → empty.

## Why it matters

The card is deliberately once-per-session (see the `hermes.ts:9-16` docblock and decision `9d327f381ede4881` on prompt-scoped recall — this is not a request to make it per-turn). But compaction is the one event that erases the card from the model's context *without* starting a new session, so once-per-session and once-per-context diverge exactly there. The mid-turn drift reminder (`host-lifecycle.ts:1334`) is the only engine output that can reach a compacted session, and only on hosts with `midTurnDeliveryVerified`.

[...]

Issue #289, opened 2026-09-09
Correction-lesson capture has never fired in production: the prompt event runs agent-reminder, which never calls the classifier


dat999zx:
## Summary

`detectCorrectionSignal` (`src/cli/agents/host-hook.ts:380`) classifies the user's prompt on `turn-start` and `recordCorrectionLesson` (`src/session/host-lifecycle.ts:1036-1046`) pends a lesson from it. The engine path works when hand-driven: piping a correction-shaped `UserPromptSubmit` into `agent-hook claude` under `capture.events=enforce` writes a `kind='correction'` row and emits the `KNOWL LESSON` nudge.

No real host has ever reached that path. Three independent gates are all closed, and no test would go red if the classifier were deleted.

## Gate 1 — config (deliberate)

`capture.events` is absent from `DEFAULT_CONFIG` (`src/core/types.ts:474-478`) and `captureEventsMode` returns `'off'` for missing/unknown (`src/store/capture-config.ts:35-38`). Only `knowl posture maximal` or `config set capture.events shadow|enforce` arms it. This is by design and not the bug — but it means every other gate has been invisible.

## Gate 2 — routing on file-configured hosts (the real defect)

claude, codex, copilot and openhands register `agent-reminder <host>` under their prompt event, **not** `agent-hook`:

- `src/cli/agents/hook-config.ts:298-311` writes `reminderEntry` under `promptEvent` and strips any lifecycle handler there
- `src/session/hosts/codex.ts:29-33` documents that the prompt event must not also be a lifecycle event
- `src/cli/agents/reminder.ts:73-115` reads `payload.prompt` only for `fleetTurnStartBestEffort` and never calls `normalizeHostHook`

So `host-lifecycle.ts`'s `turn-start` branch (1028-1093) never executes on these hosts. Verified from generated `.claude/settings.local.json`, `.codex/hooks.json`, `.github/hooks/knowl.json`.

Side effect worth noting: #275's claim that the card is "first-turn only" on all hosts is inaccurate for these four — on them the card is SessionStart-only, and `turn-start` is never seen at all.

## Gate 3 — field delivery on plugin hosts

Hosts whose turn-start *does* reach `agent-hook` never deliver root `prompt`:

- hermes — sent as `user_message` under `extra`, dropped by the allowlist (fixed in the #275 PR)
- cline — `integrations/cline/knowl-plugin.mjs:157-158` sends no prompt field
[...]

PR #291, opened 2026-09-09
fix(hermes): hoist the root keys the engine reads out of `extra`, and stop calling the .cmd shim

### LINKED ISSUE #275: Turn-start recall is recency-ordered and first-turn-only on every host; Hermes also drops the prompt at the stdin allowlist
## Summary

In a Knowl project, the automatic turn-start card is **recency-ordered and first-turn-only**. It is never scored against what the user actually asked. The only code path that does query-relevant recall is the branch for folders that are *not* Knowl projects — so the global personal-defaults store gets better automatic recall than a real project does.

This is host-independent (it lives in `host-lifecycle.ts`), but it is most visible on Hermes, where the plugin already collects the user's message and then has nowhere to send it.

## Evidence

**1. Context is only included on the first turn of a session.**

`src/session/host-lifecycle.ts:1083`

```ts
const started = await bootstrapWithHandoff(projectId, input, 'turn', !sessionBinding, turnBudget);
```

`includeContext` is `!sessionBinding`. Once the session is bound (turn 2 onward), `bootstrapAgentSession` returns at `src/store/context-bootstrap.ts:83` with `context: undefined`. Every later turn gets no card at all.

**2. The card that *is* produced is recency-based, not query-based.**

`src/core/format.ts:236` — `# KNOWL - RECENT SESSION CONTEXT`, composed from recent session activity and recent active items. Nothing in `formatRecentContext` takes a query string.

**3. The user's message reaches the plugin and dies there.**

`integrations/hermes/knowl/__init__.py:776` forwards it:

```python
user_message=str(user_message or "")[:4000],
```

It travels under `extra` in `_payload()` (line 581), and `extra` is dropped by the stdin allowlist. Even if it survived, `normalizeHostHook` reads only `raw.prompt`, and only to derive a boolean:

`src/cli/agents/host-hook.ts:380`

```ts
const correctionSignal = event === 'turn-start'
  && typeof raw.prompt === 'string' && detectCorrectionSignal(raw.prompt);
```

No retrieval is performed with it.

**4. The non-project branch does it correctly.**

`integrations/hermes/knowl/__init__.py:819`

```python
items = runner.query(text, cwd, limit=5) if (cwd and text) else []
```

A session with no project queries the global store with the user's actual words and renders a scored card. A session *with* a project does not.

## Impact

The agent's `AGENTS.md` rules say to call `knowl_query` before answering a project question. In practice the agent frequently answers straight from the recency card because a card is present and looks like recall. When that card is topically unrelated (which it usually is, being recency-ordered), the agent answers from workspace memory that has nothing to do with the question — or from nothing at all on turns 2+.

Observed live in a Hermes Desktop session on this repo: three consecutive turns asking about Knowl's positioning and team composition. The injected card held the three most recent items (an MCP stdio defect, an ambient-store rebinding fix, an OpenClaw hook finding) — none related. An explicit `knowl_query` on the same subjects returned the correct items immediately, so retrieval quality is fine; only the automatic channel is wrong.

## Proposed fix

Score the turn-start card against the user's message rather than by recency, and emit it on every turn instead of only the first:

1. Add `prompt` to the stdin allowlist for `turn-start` (already parsed at `host-hook.ts:381`, so no new field reaches storage — it stays derived-only downstream).
2. In the `turn-start` branch of `host-lifecycle.ts`, when a prompt is present, compose the card from a query against it instead of from recency.
3. Keep the recency card for the first turn / no-prompt case, which is what it is genuinely good at (orientation).

Storage promise is unaffected: the prompt is used to retrieve and then discarded, exactly as `detectCorrectionSignal` already uses it.

## Notes

Not to be confused with the separate finding that OpenClaw does not dispatch `before_prompt_build` on its embedded agent runner. That is a delivery failure on one host; this is a relevance failure on all of them.

dat999zx:
## What

Triage of #275, verified against source and reproduced on `dist/index.js`. Of the three claims, two are by design and one is real — and the real one is much larger than the issue says.

### Not fixed, by design

- **A — query-scored recall.** The proposed fix is `knowl query <user's sentence>` at turn-start. That is exactly what the Hermes `MemoryProvider.prefetch` did until `3f02b09`, removed on measurement: keyword search over prose retrieves badly, and a card that looks like an answer suppresses the agent's own `knowl_query` (6/6 → 1/6, p=0.008). Decision `9d327f381ede4881`, constraint `6676fd41b5dc410c` rule 2.
- **B — first-turn-only card.** Deliberate; `hermes.ts:9-16` documents why. Also inaccurate for claude/codex/copilot/openhands, whose prompt event runs `agent-reminder` and never reaches `turn-start` at all (#289).

### Fixed here

**C, widened.** `_payload()` nests everything under `extra`, which the stdin allowlist drops whole. Verified with `readLifecyclePayloadObject` on real payloads — hermes kept `[session_id, cwd]` and nothing else. Four root-read fields were sent, dropped, and read as absent:

| field | event | consequence |
|---|---|---|
| `prompt` | pre_llm_call | correction classifier never fired on Hermes |
| `status` / `error` | post_tool_call | a failed `write_file`/`patch`/`terminal` was captured as a **successful checkpoint**: read-set recorded, impact run on a write that didn't happen, terminal failures stored as `exitCode 0`, fleet error-signature never populated |
| `completed` / `interrupted` | on_session_end | every Ctrl-C'd turn closed as a clean `finished`; `finalizeFailedStop` never ran |
| `last_assistant_message` | pre_verify | fleet per-turn summary always null for Hermes |

Plus a fourth defect found while trying to store these findings: **`knowl_store` from the plugin fails on Windows for any multi-line content.** `_resolve_knowl_command` returned npm's `knowl.cmd`, which runs through cmd.exe and re-parses argv — a newline truncates the line (`required option '--category'`), a `|` executes. It now calls `node <dist/index.js>` directly, which is all the shim did. Reproduced before and after against a throwaway project.

### Engine side

`toolEvent` turns a root `status: 'failed'` into an error event on any host — same outcome `PostToolUseFailure` gets by event name — and a failed shell call keeps its command fingerprint. OpenClaw was already sending this pair (`integrations/openclaw/src/index.ts:380`) and silently taking the checkpoint path too.

### Tests

Every existing Hermes engine test bypassed the allowlist — `normalizeHostHook` was called with `extra` intact, so a field could be nested on either side and the suite stayed green. `tests/cli/hermes-plugin.test.ts` now routes every payload through `readLifecyclePayloadObject` first, with one case per hoisted field. `test_plugin.py` asserts each key at the root and covers the shim unwrap. The `hermes-profile.test.ts` case that pinned the `turn_id` loss as expected is left alone: Hermes does not send `turn_id` on `pre_verify`, so keying turns on it would split one turn across two bindings.

Verified: build, typecheck, lint clean; 421 files / 3959 tests pass; 52/52 Python.

Follow-ups filed: #289 (correction capture dead on all hosts — routing), #290 (no card after compaction).

Closes #275.

### DIFF
--- a/integrations/hermes/knowl/__init__.py
+++ b/integrations/hermes/knowl/__init__.py
@@ -229,7 +229,10 @@ def _resolve_knowl_command(ctx: Any) -> List[str]:
     """Where the knowl CLI is: plugin setting, then env, then PATH, then npx."""
     configured = _setting(ctx, "knowl_bin") or os.environ.get("KNOWL_BIN")
     if configured:
-        return [str(configured)]
+        # Unwrapped too, and this is the branch that matters: the documented way to point at a
+        # global install is `knowl_bin: .../npm/knowl.cmd`, so the configured path is MORE
+        # likely to be a batch shim than a discovered one, not less.
+        return _unwrap_batch_shim(str(configured))
     for candidate in ("knowl.cmd", "knowl") if sys.platform == "win32" else ("knowl",):
         found = shutil.which(candidate)
         if found:
--- a/integrations/hermes/knowl/__init__.py
+++ b/integrations/hermes/knowl/__init__.py
@@ -233,8 +233,26 @@ def _resolve_knowl_command(ctx: Any) -> List[str]:
     for candidate in ("knowl.cmd", "knowl") if sys.platform == "win32" else ("knowl",):
         found = shutil.which(candidate)
         if found:
-            return [found]
+            return _unwrap_batch_shim(found)
     return ["npx", "-y", "@dat999zx/knowl"]
+
+
+def _unwrap_batch_shim(found: str) -> List[str]:
+    """`node <dist/index.js>` in place of npm's `.cmd` shim, so no argument crosses cmd.exe.
+
+    A `.cmd` target is run through cmd.exe, which re-parses the argument line: a newline in an
+    atom's body cut the line short and `knowl store` died on a "missing" `--category`, and a
+    `|` ran the rest as a command. Quoting does not help a batch target (cmd parses twice).
+    The shim's own body is `node "%dp0%\\node_modules\\@dat999zx\\knowl\\dist\\index.js" %*`,
+    so calling that entry point directly is what the shim would have done, minus the shell.
+    """
+    if not found.lower().endswith(".cmd"):
+        return [found]
+    entry = os.path.join(os.path.dirname(found), "node_modules", "@dat999zx", "knowl", "dist", "index.js")
+    node = shutil.which("node")
+    if node and os.path.isfile(entry):
+        return [node, entry]
+    return [found]
 
 
 def _hermes_hook_callback_timeout() -> float:
@@ -565,6 +583,14 @@ def _loaded_as_memory_provider() -> bool:
     return __name__.split(".", 1)[0] == MEMORY_LOADER_NAMESPACE
 
 
+# Keys the engine reads at the ROOT of a hook payload. Everything else goes under `extra`,
+# which the engine's stdin allowlist drops whole -- so a field the engine reads has to be
+# named here or it is sent and silently lost. Each of these was: `prompt` fed the correction
+# classifier that had never fired on this host, `status`/`error` turned a failed write into a
+# recorded success, and `last_assistant_message` left the fleet's per-turn summary empty.
+_ROOT_KEYS = ("prompt", "status", "error", "last_assistant_message")
+
+
 def _payload(event: str, session_id: str, cwd: str, **extra: Any) -> Dict[str, Any]:
     """The shape Hermes' own shell-hook serializer emits (agent/shell_hooks.py:_serialize_payload)."""
     body: Dict[str, Any] = {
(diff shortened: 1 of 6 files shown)


dat999zx:
Reviewed the whole diff and re-verified the chain end to end. Both engine-side claims hold: root `status: 'failed'` from a plugin host reaches `finalizeFailedStop` via `host-hook.ts:428`, and the new tool-event branch returns the same shape `PostToolUseFailure` does (`:443-450`), so nothing downstream has to learn a second dialect. `sharesSessionBinding: true` on hermes is what makes the turn-stop path reachable, and it is set.

**One defect found and fixed on the branch (d156ea5): the unwrap only covered the PATH lookup.** `_resolve_knowl_command` returns early for a configured `knowl_bin` — and configured is the branch that matters, because `knowl_bin: .../npm/knowl.cmd` is exactly how you point the plugin at a global install. Verified against this machine's own config, which is that setup:

```
configured   -> ['C:/Users/Admin/AppData/Roaming/npm/knowl.cmd']    # still the shim
_unwrap(shim)-> ['...node.EXE', '...\\@dat999zx\\knowl\\dist\\index.js']
```

So every multi-line `knowl store` stayed broken for the documented configuration. One line, plus a test that drives `_resolve_knowl_command` with a configured shim; restoring `[str(configured)]` fails it.

The existing `_unwrap_batch_shim` tests exercise the helper directly, which is why this slipped through — the helper was right and its only guarded caller was the wrong one.

Gates on d156ea5: build, typecheck, lint, docs:check clean; 8 files / 231 vitest; 53/53 Python.

Two notes, neither blocking:
- `turn_id` left out of hermes identity is correctly argued in the docblock — `pre_verify` genuinely has no such kwarg, so keying on it would split a turn.
- `changed_paths` staying under `extra` is right: no turn-stop reader takes it.

Merging.

Issue #293, opened 2026-09-09
Windows: agent-hook writes its output, then dies with 0xC0000005 on exit


dat999zx:
Five failures across four branches and main since 2026-09-05, all Windows-only, all the same shape: a spawned `dist/index.js agent-hook claude ...` writes **correct output** and then dies with `0xC0000005` (`status: 3221225477`), no stderr.

The give-away is in the last one (run 34341310801):

```
Serialized Error: { status: 3221225477, signal: null, output: [ null,
  '{"hookSpecificOutput":{"hookEventName":"PostToolUse","additionalContext":"KNOWL CONTINUATION: ..."}}'
```

stdout is complete and correct. The hook did its whole job. The crash is on the way out, after the last `console.log`.

**Occurrences** (different test cases in the same file each time, which is what makes it look like a flake rather than one broken test):
- run 33945656223, `main` @ 5.21.1 release push — `reminders.driftEvery 0 silences the reminder`
- run 34308771583, `fix/openclaw-card-and-path` — `emits PostToolUse context after 12 non-Knowl successes and resets after Stop`
- run 34337568182, `fix/openclaw-session-and-validation` — `plugin-export` hook timeout (different symptom, same runner)
- run 34341310801, `fix/audit-store` — `reminders.driftEvery 24 moves the reminder to the 24th event`

Passes 3/3 consecutively on a local Windows box, so it is load- or timing-dependent on the runner.

**Where I would look.** `closeEntry` (`src/store/connection-pool.ts:177-186`) runs `PRAGMA wal_checkpoint(TRUNCATE)` and then `client.close()` — and `close()` is not awaited, because libSQL's is synchronous. An access violation with no JS stack, after all JS output has flushed, is the signature of the native addon's finalizer running against a handle the checkpoint is still using as the process tears down. The hook path is the only one that reliably closes a store and exits immediately; long-lived commands never hit the race.

Not blocking anything today — every occurrence has passed on re-run — but it is a native crash in the code path that runs on **every single tool call** for a Claude Code user, and on a slower machine than a CI runner it will not stay a CI-only story. If a hook dies at exit on a host that reads the exit code, `refusesOnAnyNonZeroExit` turns it into a denied edit.

Repro attempts worth making: run `tests/cli/claude-continuation-reminder.test.ts` on Windows under load (`--pool=threads --maxWorkers=N`), and check whether removing the `wal_checkpoint` from `closeEntry` on the hook path makes it go away.

[...]

PR #299, opened 2026-09-10
test(ranking+ci): the small-peer knife edge, and budgets for the slowest runner

dat999zx:
Root-causes the macOS half of #293. **It is not the same bug as the Windows crashes** — see the issue for that split — but it is fully explained, reproduced, and fixed here.

## The number was the evidence

```
expected 0.9962393332823878 to be close to 1
```

That is not float noise. It is exactly:

```
raw("Build output cleaning") / raw("Deploy build output")
  = 3.846264921786819e-6 / 3.860784043844392e-6
  = 0.9962393332823880          (CI: ...878)
```

So the *other* peer row led the page. Both rows fully cover the query; their raw lexical evidence differs by **0.4%**.

## Why a machine can decide it

`normalizedRecencyScore` (`agent-query.ts:219-232`) is min-max over the candidate set:

```
recency = (row.updatedAt - oldest) / (newest - oldest)
```

That is **scale-free** — a 1ms gap is worth a full point of recency when it is the only gap in the set, and nearly nothing when the set spans a minute. `Build output cleaning` is seeded last, so it is the newest row. How much that newness is worth therefore depends on the total span the fixture happens to cover, which is decided by *how fast the machine wrote the rows*, not by anything in the ranking.

Reproduced by holding the ranking fixed and varying only that span. Same two rows, same lexical gap, different winner.

## What was actually wrong: the assertion, not the code

The test's real claim survives the failure. `expect(scored[0].repo).toBe('small')` — the clamped tiny peer still wins the page, which is the entire point of the test — **passes in the failing case too**. I verified this explicitly (`peerLeadsPage=true` in every configuration, including the one that produces the CI number).

Only `toBeCloseTo(1, 5)` broke, and it was asserting something the test never meant: that of two indistinguishable peer rows, one specific one leads. Lexical is a tie-breaker rather than a veto here (`FUSION_ALPHA`), and a 0.4% gap is a tie. Both orderings are correct.

So it now asserts the recovery (`> 0.99`) rather than the tie-break. **Still strict** — reverting the per-corpus normalisation this test exists to guard fails 4 of its 5 cases.

Second file pins the mechanism, so the property is stated deliberately rather than rediscovered as a flake next time.

Gates: typecheck, lint clean; `tests/store` 144 files / 1178 tests green.

Refs #293 — the Windows access violations and stalls are a separate, still-unreproduced problem and this does not touch them.

### DIFF
--- a/tests/benchmark/knowl-adapter.test.ts
+++ b/tests/benchmark/knowl-adapter.test.ts
@@ -21,7 +21,9 @@ const records: NormalizedRecord[] = [
   },
 ];
 
-describe('Knowl normalized benchmark adapter', () => {
+// Spawns the real CLI per case; timed out at 32.4s on windows-latest against the 30s default.
+// See #293 -- the Windows runner is ~2x slower than ubuntu on this suite's workload.
+describe('Knowl normalized benchmark adapter', { timeout: 120_000 }, () => {
   it('preserves source IDs, applies supersession, and isolates projects', async () => {
     const adapter = new KnowlBenchmarkAdapter();
     await adapter.reset({ runId: 'knowl-adapter-test', mode: 'normalized', seed: 1 });
--- a/tests/cli/temporal-cli.test.ts
+++ b/tests/cli/temporal-cli.test.ts
@@ -20,7 +20,10 @@ describe('temporal CLI', () => {
     const client = createClient({ url: `file:${path.join(ROOT, '.knowl', 'knowl.db')}` });
     const row = (await client.execute('SELECT id, created_at FROM knowledge_items WHERE title = ?', ['Temporal storage'])).rows[0];
     itemId = String(row.id); createdAt = String(row.created_at); client.close();
-  }, 15_000);
+  // Two cold CLI spawns, on a budget tighter than the suite default. ~300ms each locally and
+  // roughly double on the Windows runner, so this had the least headroom in the repo and is
+  // where a slow runner hit first (#293).
+  }, 120_000);
   afterAll(async () => { await fs.rm(ROOT, { recursive: true, force: true }).catch(() => {}); });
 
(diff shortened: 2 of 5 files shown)


dat999zx:
Pushed a second commit (fb8b4f4) fixing the **timeout** half, now that it is measured rather than guessed.

Correcting myself on the way: my first estimate blamed `claude-continuation-reminder`'s 24-iteration loops, but those already carry 120–180s budgets. The real victims were suites sitting on the **default** 30s (or tighter):

| suite | local | budget | observed on Windows |
|---|---|---|---|
| `namespace-id-resolution` | ~12s | 30s (default) | **33.3s**, **30.6s** |
| `knowl-adapter` | — | 30s (default) | **32.4s** |
| `temporal-cli` | — | **15s** (tighter than the default it overrode) | hook timeout |

Under 3× of headroom against a runner that is ~2× slower (7.0min vs ubuntu's 3.7min wall on the same commit). All three now get 120s — 10× the measured local cost, not a number chosen to make CI quiet.

**Deliberately left alone: `plugin-export`.** Its 300s hook also timed out twice, but it costs **38s locally** — already 8× headroom. That failure is `npm pack` + a real `npm install` hitting the registry, and a bigger number would only lengthen the wait before the same failure.

## Where #293 stands

- **macOS scoring drift** → #299, root-caused to a scale-free recency prior making a 0.4% lexical gap decidable by fixture-write speed. Fixed, with the mechanism pinned by its own test.
- **Windows timeouts** → fb8b4f4, above. Understood and fixed.
- **`0xC0000005`** → still unexplained, and I am not shipping a guess. 388 failed reproduction attempts; my two hypotheses (floating promises in `closeEntry`, shared-`KNOWL_HOME` contention) are both disproved above.

Once these land, this issue should be **only** the access violation. Everything else filed under it has a cause and a fix.

